PR #52, opened 2015-02-18
Allow to set resolvers via web.config/app.config

steven-r:
Currently there's no chance to replace the given ISubmissionClient resolver without changing the source code of exceptionless when using ExceptionlessClient. This change allows to set resolvers via web.config/app.config with code located outside of exceptionless code.


steven-r:
hmm. The problem might be that it will become hard to find a generic way to differentiate between the standard setting (see RegisterDefaultServices()) and any custom setting. Maybe the default should be null (this should work as DependencyResolverExtensions check always if the value is null - in this case the default resolver is returned).

The same check could be done in Startup() (if RegisterDefaultServices() is updated).

Issue #57, opened 2015-02-26
Silverlight Support


niemyjski:
I noticed that there are some current issues with the PCL client when using Silverlight 5. If you get the following errors this is how you get around them:

				TinyIoC.TinyIoCResolutionException occurred
	  Message=Unable to resolve type: Exceptionless.Enrichments.Default.ConfigurationDefaultsEnrichment
	  StackTrace:
		   at TinyIoC.TinyIoCContainer.ConstructType(Type requestedType, Type implementationType, ConstructorInfo constructor, NamedParameterOverloads parameters, ResolveOptions options)
	  InnerException: System.MethodAccessException
		   Message=Attempt by method 'System.Linq.Expressions.Expression.CreateLambda(System.Type, System.Linq.Expressions.Expression, System.String, Boolean, System.Collections.ObjectModel.ReadOnlyCollection`1<System.Linq.Expressions.ParameterExpression>)' to access method 'System.Linq.Expressions.Expression`1<TinyIoC.TinyIoCContainer+ObjectConstructor>.Create(System.Linq.Expressions.Expression, System.String, Boolean, System.Collections.ObjectModel.ReadOnlyCollection`1<System.Linq.Expressions.ParameterExpression>)' failed.
		   StackTrace:
				at System.RuntimeMethodHandle.PerformSecurityCheck(Object obj, RuntimeMethodHandleInternal method, RuntimeType parent, UInt32 invocationFlags)
				at System.RuntimeMethodHandle.PerformSecurityCheck(Object obj, IRuntimeMethodInfo method, RuntimeType parent, UInt32 invocationFlags)
				at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture, Boolean skipVisibilityChecks)
				at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
				at System.Linq.Expressions.Expression.CreateLambda(Type delegateType, Expression body, String name, Boolean tailCall, ReadOnlyCollection`1 parameters)
				at System.Linq.Expressions.Expression.Lambda(Type delegateType, Expression body, String name, Boolean tailCall, IEnumerable`1 parameters)
				at System.Linq.Expressions.Expression.Lambda(Type delegateType, Expression body, ParameterExpression[] parameters)
[...]

PR #59, opened 2015-02-27
Fixes for Silverlight

jaddington:
Blake,

Here you go... seems to do the trick.  I didn't look into the Elevated Permissions bit..

Must close stream for AsyncWaitHandle for a HttpWebRequest (bombs in SL)
but good practice.

The HttpWebRequest for SL does not have a UserAgent property so must set
via headers.

### DIFF
--- a/Source/Clients/Shared/Extensions/WebRequestExtensions.cs
+++ b/Source/Clients/Shared/Extensions/WebRequestExtensions.cs
@@ -42,9 +42,13 @@ namespace Exceptionless.Extensions {
         private static readonly Lazy<PropertyInfo> _userAgentProperty = new Lazy<PropertyInfo>(() => typeof(HttpWebRequest).GetProperty("UserAgent"));
 
         public static void SetUserAgent(this HttpWebRequest request, ExceptionlessConfiguration configuration) {
+
             if (_userAgentProperty.Value != null) {
                 try {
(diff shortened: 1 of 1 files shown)


jaddington:
okay updated... not sure how you want to do the UserAgent otherwise.  I prefer not to let it fall into that second exception but it works...  Feel free to tweak.  

CLA signed.

niemyjski:
Thanks for your commit. I ended up removing the NotSupportedException catch. When I debugged this this was a TargetInvocationException that
wrapped a NotSupportedException. We know how to parse the bottom header
so all works :) Thanks man!

niemyjski:
This fixes one of the issues in #57

jaddington:
Great.  Glad I could help.  I’ll look for the updated nuget packages and then plug it in.

niemyjski:
They can be found on myget (which is currently experiencing some issues). We haven't pushed the latest ones yet

jaddington:
I grabbed the 1166 package from MyGet and plugged it in.  Can’t tell if its working, the .io site is offline.  I’ll take a peek later this afternoon or if I hear from you.

Thx

jack

niemyjski:
Yeah, we are currently migration data.

niemyjski:
The preview site is online again and we are working on pushing updated preview packages.

PR #62, opened 2015-03-12
Using await instead of waiting for an async method

kernel32uiuc:


niemyjski:
Nice catch! Can you please sign our cla: http://www.clahub.com/agreements/exceptionless/Exceptionless so we can accept this pull request

kernel32uiuc:
Sure, I did.

niemyjski:
Thanks!

PR #120, opened 2015-07-13
Removed runAllManagedModulesForAllRequests setting for performance reasons

niemyjski:
http://www.hanselman.com/blog/BackToBasicsDynamicImageGenerationASPNETControllersRoutingIHttpHandlersAndRunAllManagedModulesForAllRequests.aspx

@ejsmith  I tried looking back in history and figuring out why this was added but
couldn't find anything other than it was added as a generic commit in 2012.  I did a really quick test and nothing seems broken.


niemyjski:
@ejsmith I'm accepting this.

PR #155, opened 2015-11-10
Trying to fix server side of: https://github.com/exceptionless/Except…

modesto:
It seems that the API to create the project already gets the OrganizationId but the check-name api is calling IsProjectNameAvailableInternalAsync method with this value as null.

I'm not sure how to change UI side to call chec-name API with OrganizationId param. Could someone help with this part?


niemyjski:
Thank you so much for your pull request!

Issue #147, opened 2015-10-13
Add setting to allow InviteToken to override EnableAccountCreation


theit8514:
Currently, EnableAccountCreation disables all account creations, even those invited to an organization with a valid InviteToken. Was looking to see if a configuration setting could be added to allow InviteTokens to be used. In my situation, I might have several organizations for different teams in my self-hosted instance and would need to go to the web.config and modify EnableAccountCreation any time one of those organizations need to add a coworker. With this settings, they can use the current Invite system to invite people to organizations while still preventing users outside our organization from creating accounts.

Some code and tests written already: https://github.com/theit8514/Exceptionless/commit/ddb1721ed7abcd44c06a88ffa113367bdeabfb31. Existing functionality is tested as well. Both settings off result in no accounts being accepted, and accounts should be accepted when EnableAccountCreation is enabled regardless of the InviteToken setting.

niemyjski:
This would be good to have. I can see people wanting this scenario. It should be pretty easy to validate the token inside of the controller action and bypass that check. thoughts?

niemyjski:
@theit8514 would you mind submitting a pull request for this.

niemyjski:
Closed via https://github.com/exceptionless/Exceptionless/pull/174

PR #174, opened 2016-01-15
Add setting to allow InviteToken to override EnableAccountCreation

theit8514:
Adds logic to SignupAsync to allow InviteTokens to override the EnableAccountCreation flag when the EnableInvitedUserCreation flag is set. Requested PR from issue #147.

Unit tests are modeled after existing controller tests but are flaky when run as a group because of xUnit's async task runner and the updating of the global Settings. Let me know if you want to trim this down to just the logic change.

### DIFF
--- a/Source/Api/Controllers/AuthController.cs
+++ b/Source/Api/Controllers/AuthController.cs
@@ -116,7 +116,8 @@ namespace Exceptionless.Api.Controllers {
         [Route("signup")]
         [ResponseType(typeof(TokenResult))]
         public async Task<IHttpActionResult> SignupAsync(SignupModel model) {
-            if (!Settings.Current.EnableAccountCreation)
+            var valid = await IsAccountCreationEnabledAsync(model.InviteToken);
+            if (!valid)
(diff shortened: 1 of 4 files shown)


ejsmith:
Would it be better to just always allow invite tokens to work even if account creation is turned off?

theit8514:
Adding the token to the check there wasn't working. I split the two out.

theit8514:
I'm okay with always allowing Invite Tokens to always work.

niemyjski:
Thanks for your work! If the tests are flakey I'd rather not introduce them as we already have a few rouge unit tests I've been trying to track down. 

I'd also prefer just using the token if it's valid to override the account creation process. This would also keep the amount of settings we have to a minimum.

theit8514:
I've cleaned up the setting and squashed all the commits. The concurrency issue I was having was related to RemoveAllAsync on the users repository. With each test getting it's own email, they no longer interfere with each other.

niemyjski:
Thank you!

PR #175, opened 2016-01-18
Updates to Sessions

niemyjski:


Source/Core/Extensions/PersistentEventExtensions.cs
```diff
@@ -105,6 +152,11 @@ public static class PersistentEventExtensions {
                 Type = Event.KnownTypes.Session,
                 Value = 0
             };
+            
+            startEvent.SetSessionId(source.GetSessionId());
```
niemyjski: This might throw a null reference exception

Issue #178, opened 2016-01-21
Flow the client ip address from event submission into the pipeline context and events request info


niemyjski:
We should append this ip address to the request info client ip address list.

teamaton:
We would like to see the IP address on the Request tab of an Event Occurrence and also in the Session, where a very similar looking tab is named Browser.

niemyjski:
This should add the ip if it's not already there, but the request tab is the browser tab on session events :)

PR #179, opened 2016-01-21
#178 Added the ability to pass the clients submission ip address into the events request info

niemyjski:


Source/Core/Plugins/EventProcessor/Default/40_RequestInfoPlugin.cs
```diff
@@ -40,36 +40,51 @@ public class RequestInfoPlugin : EventProcessorPluginBase {
...
+        }
+
+        private void SetClientIPAddress(RequestInfo request, string clientIPAddress) {
+            if (String.IsNullOrEmpty(clientIPAddress))
+                return;
 
-                    if (!String.Equals(info.Device.Family, "Other"))
-                        request.Data[RequestInfo.KnownDataKeys.Device] = info.Device.Family;
+            if (clientIPAddress.IsLocalHost())
+                return;
```
niemyjski: We are returning early if the ip address from the submission request is local host. I'm not 100% sure if want to do this but it would cut down on us having to check for both `::1` and `127.0.0.1`.

Source/Core/Plugins/EventProcessor/EventContext.cs
```diff
@@ -3,15 +3,18 @@
 using Exceptionless.Core.Pipeline;
 using Exceptionless.Core.Utility;
 using Exceptionless.Core.Models;
+using Exceptionless.Core.Queues.Models;
 
 namespace Exceptionless.Core.Plugins.EventProcessor {
     public class EventContext : ExtensibleObject, IPipelineContext {
-        public EventContext(PersistentEvent ev) {
+        public EventContext(PersistentEvent ev, EventPostInfo epi = null) {
```
niemyjski: Kinda stinks having our event context dependent upon the post info but I could see you wanting to do stuff with it..

Issue #180, opened 2016-01-23
Custom Stacking Behavior


adamzolotarev:
It would be useful to have some sort of property on the event that you can set that would override the default stacking behaviour and let you stack things in custom ways.

Currently, as a workaround, if you want to override default exception stacking behavior, you can do this: 
```csharp
ExceptionlessClient.Default.CreateEvent().SetType(Event.KnownTypes.Error).SetSource("customSource");
```
And "customSource" will serve as a new stacking. But then you lose the Exception tab and you have to manually translate the Exception information.

But if instead you do
```csharp
exception.ToExceptionless().SetSource("customSource"); 
```
it still uses a default stacking behavior (exception type + method).

Not sure if it would be better to add a completely new method to set custom grouping (e.g., SetCustomStack("customStackName)") or just add a property (e.g., IsCustomGrouping) and just use the Source property to group the exceptions.

niemyjski:
We had a quick talk about this over the weekend and we like this idea, would you be willing to take a stab at implementing this in a pull request?

Here is some of our thoughts on this:
* Create EventBuilder and Event extension methods on the client to handle this, but place them under a different extensions namespace (maybe Exceptionless.Advanced??). The reasoning for putting them in a different namespace is we want to make it crystal clear that this is something advanced and not a normal scenario and you understand what you are doing. This extension method should set an extended property on the event and call it something like `@StackSignature` with a string value that you want to stack on. The extension method that we create should validate that this value is set. 
* Update the [pipeline/AssignToStack pipeline action](https://github.com/exceptionless/Exceptionless/blob/master/Source/Core/Pipeline/010_AssignToStackAction.cs#L40) to use this value for the `StackSignatureData`. *Optional/Feedback* Should we remove this extended property from the event once we use it on signature data?

niemyjski:
@ejsmith can you provide any additional feedback?

ejsmith:
So 1st off, the fact that we aren't showing the error tab when there is an error object on the event is a bug. Event type shouldn't matter for showing that tab. Only that the data is present.

I am definitely on board with allowing custom stacking and agree that I am a little worried that people would see something like SetStackingKey and think they needed to set it. But I think if we do something like SetManualStackingKey then I think it is obvious enough that we can just include it with the other extension methods.  We would also want a good XML doc description saying what the method does.

- Add SetManualStackingKey(string stackingKey) that sets an extended property called @StackingKey (need to verify that this name matches our other @ names).
[...]

PR #185, opened 2016-01-30
Added ability to do manual stacking

adamzolotarev:
When a client sends a ManualStackingKey in the Event.Data, use it to define custom stacking behavior.
For issue #180

### DIFF
--- a/Source/Core/Models/Event.cs
+++ b/Source/Core/Models/Event.cs
@@ -81,6 +81,7 @@ namespace Exceptionless.Core.Models {
             public const string SubmissionMethod = "@submission_method";
             public const string SessionEnd = "sessionend";
             public const string SessionHasError = "haserror";
+            public const string ManualStackingKey = "@manual_stacking_key";
         }
     }
 }
--- a/Source/Core/Plugins/EventProcessor/Default/20_ErrorPlugin.cs
+++ b/Source/Core/Plugins/EventProcessor/Default/20_ErrorPlugin.cs
@@ -9,8 +9,7 @@ using Exceptionless.Core.Models.Data;
 namespace Exceptionless.Core.Plugins.EventProcessor {
     [Priority(20)]
     public class ErrorPlugin : EventProcessorPluginBase {
-        public override Task EventProcessingAsync(EventContext context)
-        {
+        public override Task EventProcessingAsync(EventContext context) {
             if (!context.Event.IsError())
                 return Task.CompletedTask;
 
@@ -19,7 +18,7 @@ namespace Exceptionless.Core.Plugins.EventProcessor {
                 return Task.CompletedTask;
 
             if (String.IsNullOrWhiteSpace(context.Event.Message))
-                context.Event.Message = error.Message;
+                context.Event.Message = error.Message;            
 
             string[] commonUserMethods = { "DataContext.SubmitChanges", "Entities.SaveChanges" };
             if (context.HasProperty("CommonMethods"))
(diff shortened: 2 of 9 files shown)


Source/Core/Models/Event.cs
```diff
@@ -81,6 +81,7 @@ public static class KnownDataKeys {
             public const string SubmissionMethod = "@submission_method";
             public const string SessionEnd = "sessionend";
             public const string SessionHasError = "haserror";
+            public const string ManualStackingKey = "@manual_stacking_key";
```
niemyjski: @ejsmith do you like this name? I kind of like `@override_stacking_hash` or `@advanced_stacking`. Since we have live with this name for a really long time, I'd like some kind of overall consensus? @adamzolotarev thoughts?

I'm wondering what we would display to the user to show how it was stacked? What if a user wanted to stack it on a few different things.. would it be a key value pairs in a commented list or what?

ejsmith: I like it. I think it's very clear what it does.

If they want to stack on multiple things then they would need to build up their own hash based on those multiple things. I guess it would be nice to allow the value to be a simple string or an object, but I think that just adds more complexity than it's worth.

Source/Core/Plugins/EventProcessor/Default/20_ErrorPlugin.cs
```diff
@@ -9,7 +9,8 @@
 namespace Exceptionless.Core.Plugins.EventProcessor {
     [Priority(20)]
     public class ErrorPlugin : EventProcessorPluginBase {
-        public override Task EventProcessingAsync(EventContext context) {
+        public override Task EventProcessingAsync(EventContext context)
+        {
```
ejsmith: Can you please reformat this to match our formatting style? :-)

Source/Core/Plugins/EventProcessor/Default/20_ErrorPlugin.cs
```diff
@@ -28,6 +29,7 @@ public class ErrorPlugin : EventProcessorPluginBase {
             if (context.HasProperty("UserNamespaces"))
                 userNamespaces = context.GetProperty<string>("UserNamespaces").SplitAndTrim(',');
 
+            AddManualStackingKeyToSignatureIfExists(context);
```
ejsmith: I was thinking that this would be a separate step in the pipeline because it really has nothing to do with errors. It can be applied to any type of event.

adamzolotarev: I'll update the formatting - sorry, missed that. And, yeah it makes sense for it to be a separate step. I'll try to update in the next couple days, but may not get to it till this weekend. Thanks for the feedback!

adamzolotarev: I think I may need a little help on this. Are you saying I should create another plugin which shouldl run before any other plugins (maybe right after 0_ThrottleBotsPlugin)? 
Or should I add a new action to the pipeline which would run before 005_RunEventProcessingPluginsAction?

ejsmith: Yeah a new plugin that runs after throttle bots.

niemyjski:
@adamzolotarev Please let me know if you have any questions

adamzolotarev:
Think I messed up my merging. I'll have to look at it tomorrow

adamzolotarev:
I did something weird with this branch. Just created a new one and send you a new pull request.

PR #187, opened 2016-02-06
Manual stacking

adamzolotarev:
Added manual stacking plugin. Don't add signature data in the error plugin if already added manual stacking. Error plugin still sets message and target info.

### DIFF
--- a/Source/Core/Models/Event.cs
+++ b/Source/Core/Models/Event.cs
@@ -81,6 +81,7 @@ namespace Exceptionless.Core.Models {
             public const string SubmissionMethod = "@submission_method";
             public const string SessionEnd = "sessionend";
             public const string SessionHasError = "haserror";
+            public const string ManualStackingKey = "@manual_stacking_key";
         }
     }
 }
--- a/Source/Core/Models/Event.cs
+++ b/Source/Core/Models/Event.cs
@@ -81,7 +81,7 @@ namespace Exceptionless.Core.Models {
             public const string SubmissionMethod = "@submission_method";
             public const string SessionEnd = "sessionend";
             public const string SessionHasError = "haserror";
-            public const string ManualStackingKey = "@manual_stacking_key";
+            public const string ManualStackingKey = "@stack";
         }
     }
 }
--- a/Source/Core/Plugins/EventProcessor/Default/03_ManualStackingPlugin.cs
+++ b/Source/Core/Plugins/EventProcessor/Default/03_ManualStackingPlugin.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.Threading.Tasks;
 using Exceptionless.Core.Pipeline;
 using Exceptionless.Core.Extensions;
 using Exceptionless.Core.Models;
@@ -8,11 +9,8 @@ namespace Exceptionless.Core.Plugins.EventProcessor {
     public class ManualStackingPlugin : EventProcessorPluginBase {
         public override Task EventProcessingAsync(EventContext context) {
             var data = context.Event.Data;
-            if (!data.ContainsKey(Event.KnownDataKeys.ManualStackingKey)
-                || data[Event.KnownDataKeys.ManualStackingKey] == null)
-                return Task.CompletedTask;
-
-            context.StackSignatureData.AddItemIfNotEmpty(nameof(Event.KnownDataKeys.ManualStackingKey), data[Event.KnownDataKeys.ManualStackingKey].ToString());
+            if (data.ContainsKey(Event.KnownDataKeys.ManualStackingKey) && data[Event.KnownDataKeys.ManualStackingKey] != null)
+                context.StackSignatureData.AddItemIfNotEmpty(nameof(Event.KnownDataKeys.ManualStackingKey), data[Event.KnownDataKeys.ManualStackingKey].ToString());
 
             return Task.CompletedTask;
         }
--- a/Source/Core/Plugins/EventProcessor/Default/03_ManualStackingPlugin.cs
+++ b/Source/Core/Plugins/EventProcessor/Default/03_ManualStackingPlugin.cs
@@ -1,0 +1,20 @@
+﻿using System.Threading.Tasks;
+using Exceptionless.Core.Pipeline;
+using Exceptionless.Core.Extensions;
+using Exceptionless.Core.Models;
+
+namespace Exceptionless.Core.Plugins.EventProcessor {
+    [Priority(3)]
+    public class ManualStackingPlugin : EventProcessorPluginBase {
+        public override Task EventProcessingAsync(EventContext context) {
+            var data = context.Event.Data;
+            if (!data.ContainsKey(Event.KnownDataKeys.ManualStackingKey)
+                || data[Event.KnownDataKeys.ManualStackingKey] == null)
+                return Task.CompletedTask; ;
+
+            context.StackSignatureData.AddItemIfNotEmpty(nameof(Event.KnownDataKeys.ManualStackingKey), data[Event.KnownDataKeys.ManualStackingKey].ToString());
+
+            return Task.CompletedTask;
+        }
+    }
+}
--- a/Source/Core/Plugins/EventProcessor/Default/03_ManualStackingPlugin.cs
+++ b/Source/Core/Plugins/EventProcessor/Default/03_ManualStackingPlugin.cs
@@ -10,7 +10,7 @@ namespace Exceptionless.Core.Plugins.EventProcessor {
             var data = context.Event.Data;
             if (!data.ContainsKey(Event.KnownDataKeys.ManualStackingKey)
                 || data[Event.KnownDataKeys.ManualStackingKey] == null)
-                return Task.CompletedTask; ;
+                return Task.CompletedTask;
 
             context.StackSignatureData.AddItemIfNotEmpty(nameof(Event.KnownDataKeys.ManualStackingKey), data[Event.KnownDataKeys.ManualStackingKey].ToString());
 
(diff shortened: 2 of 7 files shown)


Source/Core/Plugins/EventProcessor/Default/03_ManualStackingPlugin.cs
```diff
@@ -0,0 +1,20 @@
...
+using Exceptionless.Core.Models;
+
+namespace Exceptionless.Core.Plugins.EventProcessor {
+    [Priority(3)]
+    public class ManualStackingPlugin : EventProcessorPluginBase {
+        public override Task EventProcessingAsync(EventContext context) {
+            var data = context.Event.Data;
+            if (!data.ContainsKey(Event.KnownDataKeys.ManualStackingKey)
+                || data[Event.KnownDataKeys.ManualStackingKey] == null)
+                return Task.CompletedTask; ;
```
niemyjski: There is an extra comma here, no biggie :)

Source/Core/Plugins/EventProcessor/Default/20_ErrorPlugin.cs
```diff
@@ -39,6 +39,9 @@ public class ErrorPlugin : EventProcessorPluginBase {
 
             error.Data[Error.KnownDataKeys.TargetInfo] = targetInfo;
 
+            if (context.StackSignatureData.Count > 0)
```
niemyjski: Any specific reason for this change? I'm a bit worried about it as it could change the behavior of how things were working previously? Should we be checking to see if stack signature data contains our known key here instead?

Is there any chance we could add some unit tests for this? I could even add them..

niemyjski:
Thanks so much for you pull request. I've added some comments above.

Source/Core/Plugins/EventProcessor/Default/20_ErrorPlugin.cs
adamzolotarev: If I remove this line, then the stack signature will be different. I want the signature to be based on the manual stacking key only. I think it makes sense to check just the known key instead. I'll update the code.

I'll write some tests. Thanks!

adamzolotarev: Hmm, I don't see a project for Exceptionles.Core tests, which is where I made my changes.

adamzolotarev:
Updated based on your suggestions. But, I don't see a project for Exceptionless.Core.Tests

niemyjski:
@adamzolotarev We just have a single tests project and that covers both the api and core projects. I think what you are looking for is the pipeline tests where there should be some stacking related tests.

Source/Core/Plugins/EventProcessor/Default/03_ManualStackingPlugin.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Exceptionless.Core.Plugins.EventProcessor {
+    [Priority(3)]
+    public class ManualStackingPlugin : EventProcessorPluginBase {
+        public override Task EventProcessingAsync(EventContext context) {
+            var data = context.Event.Data;
+            if (!data.ContainsKey(Event.KnownDataKeys.ManualStackingKey)
+                || data[Event.KnownDataKeys.ManualStackingKey] == null)
+                return Task.CompletedTask;
+
+            context.StackSignatureData.AddItemIfNotEmpty(nameof(Event.KnownDataKeys.ManualStackingKey), data[Event.KnownDataKeys.ManualStackingKey].ToString());
```
ejsmith: It's the key name here wrong? I think it would resolve to "ManualStackingKey" instead of "@manual_stacking_key".

Source/Core/Models/Event.cs
```diff
@@ -81,6 +81,7 @@ public static class KnownDataKeys {
             public const string SubmissionMethod = "@submission_method";
             public const string SessionEnd = "sessionend";
             public const string SessionHasError = "haserror";
+            public const string ManualStackingKey = "@manual_stacking_key";
```
ejsmith: Wondering if we should pick something simpler for this as I want it to feel close to a 1st class thing without actually being a real property on the event. Maybe just "@stack"?

Source/Core/Plugins/EventProcessor/Default/20_ErrorPlugin.cs
```diff
@@ -39,6 +39,9 @@ public class ErrorPlugin : EventProcessorPluginBase {
 
             error.Data[Error.KnownDataKeys.TargetInfo] = targetInfo;
 
+            if (context.StackSignatureData.ContainsKey(nameof(Event.KnownDataKeys.ManualStackingKey)))
```
ejsmith: I think we should really be just checking to see if StackSignatureData.Count > 0 and exiting. If stacking information has been supplied in some other way then we probably shouldn't override it.

Source/Core/Plugins/EventProcessor/Default/03_ManualStackingPlugin.cs
adamzolotarev: This is what will be displayed as the Stacking Information. So, yes, it would resolve to "ManualStackingKey". I guess what would you want to see there:
image

Source/Core/Models/Event.cs
adamzolotarev: Sure, I can rename it. I would have to submit another pool request for the .NET client, since these names have to match.

Source/Core/Plugins/EventProcessor/Default/20_ErrorPlugin.cs
adamzolotarev: Ok, I'll change it.

niemyjski:
@adamzolotarev is there anything I can help you with? If you like I could help stub out the tests if you give me write access to your repo.

niemyjski:
If you want I can update the .net client. I need to get a release pushed out for that soon with some fixes.

adamzolotarev:
I added the tests in my latest commit. I guess I just want a confirmation that you want to change the "@manual_stacking_key" to "@stack". But feel free to commit to my branch. I gave you write access. I usually only have time to make my changes in evenings or weekends

niemyjski:
@adamzolotarev `@stack` will be the new key name :)

Source/Core/Plugins/EventProcessor/Default/03_ManualStackingPlugin.cs
ejsmith: We can make this whatever we want it to be. I am trying to make this clean and simple all the way down to the raw HTTP level. So if I do a CURL command to post data, I don't want to use @manual_stacking_key, it's just too verbose and I think @stack is still pretty self explanatory.

As for the StackSignatureData, we can add it as whatever we want. So I'd say something like ManualStackingKey is good there.

Source/Core/Models/Event.cs
ejsmith: Yeah, sorry about that.

adamzolotarev:
Ok, I changed it to @stack. And I see you already changed .Net client. 
Thanks! Let me know if there is anything else I can do for this change.

Source/Core/Plugins/EventProcessor/Default/20_ErrorPlugin.cs
```diff
@@ -19,6 +19,9 @@ public class ErrorPlugin : EventProcessorPluginBase {
 
             if (String.IsNullOrWhiteSpace(context.Event.Message))
                 context.Event.Message = error.Message;
+            
+            if (context.StackSignatureData.Count > 0)
+                return Task.CompletedTask;
```
niemyjski: @ejsmith is there ever a time where we would want to run stacking data and set the target info if you have a manual stack signature?

Source/Core/Plugins/EventProcessor/Default/30_SimpleErrorPlugin.cs
```diff
@@ -17,6 +17,9 @@ public class SimpleErrorPlugin : EventProcessorPluginBase {
             
             if (String.IsNullOrWhiteSpace(context.Event.Message))
                 context.Event.Message = error.Message;
+            
+            if (context.StackSignatureData.Count > 0)
+                return Task.CompletedTask;
```
niemyjski: @adamzolotarev @ejsmith I also added this to the simple error.

ejsmith:
This is looking pretty good. Think we can get this merged in today @niemyjski ?

niemyjski:
Merged :) Thank you @adamzolotarev

adamzolotarev:
Thank you! And to your question about whether we would want to run stacking data and set the target info if you have a manual stack signature - initially, I didn't set the target info, but then the exception event would not have information about the method or the message. Don't know if there is a better way to handle it.
image

niemyjski:
@ejsmith yeah, we may want to set the target info on the event so we get nicer summaries it just extra work for data we aren't using. (as we are using the manual stacking key).

ejsmith:
At this point you've taken control of the stacking and we don't know what the target info is. I think we might have to improve on this feature and provide some way to set the stack title as well as the key. Thinking we should use the stacking key as the stack title for now. Thoughts?

adamzolotarev:
Maybe, when we call SetManualStacking,  we should pass an object (something like ManualStackingInfo) instead of a string. That would make it easier to set more properties like the stack title.
SetManualStacking(new ManualStackingInfo{Title ="Title", ManualStackingKey = "ManualKey"}

ejsmith:
Yeah, I was thinking that too. Maybe it could be a string like it is now and we will do the best we can to come up with a title. Or you can send an object that lets you give a title and a list of stacking values.

niemyjski:
Or we could just use the message property of the event with the manual stacking key and keep it simple

ejsmith:
We should fallback to the message as a title now, but that isn't going to be accurate when it's different for errors in the stack or when you want to use message for something else.

ejsmith:
Also, it would be nice to specify multiple pieces of the stack signature like we do with errors now so you can see what went into the stack signature

PR #193, opened 2016-02-16
Ability to get stats on generic fields

niemyjski:
Adds support for average, distinct, sum, min, max


niemyjski:
I noticed in some places the returned value from elastic (in buckets is null, we convert this to 0). We might want to look into a default field value for `value`. Also do we want timelines to have a first and last occurrence aggregation by default?

Source/Core/Models/Stats/NumbersTimelineStatsResult.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+
+namespace Exceptionless.Core.Models.Stats {
+    public class NumbersTimelineStatsResult : NumbersStatsResult {
+        public NumbersTimelineStatsResult() {
+            Timeline = new List<NumbersTimelineItem>();
+        }
+
+        public ICollection<NumbersTimelineItem> Timeline { get; private set; }
+        public double AvgPerHour { get; set; }
```
niemyjski: @ejsmith Wondering if these should be here.

PR #194, opened 2016-02-19
Support for third party clients

niemyjski:


niemyjski:
@skoub I just did a bunch of work on your branch and did some research and talked to @billhenning (Bill owns http://www.actiprosoftware.com/) and has a lot of experience with parsers / lexers.

I started diving into those methods and quickly realizes that the method name could contain this
`my.long.namespace.methodname<generics<that<never<end>>>(Blah<moreGenerics>(OtherFunctionName()) , 1)` as an example and all of those namespaces and types have long namespaces and we need to populate generic info as well as parameter with generic info
which is a BEATING multiplied by X Languages.

Ao I looked really quickly at antlr and that appears to be a mess and hasn’t been updated on .net in ages
I’ve come to the conclusion that it’s just better to use `SimpleError` and build out the stack traces
it’s a stop gap and gets you into our system from a platform we don’t yet support we can then work on writing a seperate api end point (service) in the future that uses java or some x language with a ton of awesome grammers to parse stack traces..

We might just do is put `Name = stackTrace.MethodName`, but that’s going to lead to formatting issues client side and maybe some other issues? Please let me know what you think but I think the `SimpleError` is the best way to go.

skoub:
I don't know what's SimpleError is doing. I will have to check that. If i understand right, we will continue to parse Raygun with our logic but use SimpleError to parse the stacktrace? Or you dump everything and only use SimpleError?

niemyjski:
Here is a link to where we populate it.. https://github.com/exceptionless/Exceptionless.Net/blob/master/Source/Shared/Extensions/ToSimpleErrorModelExtensions.cs

So in my new changes you'd comment out MapError(model) and override MapSimpleError(model). The `Stacktrace` property needs to be set which is the raw stack trace (as shown when an error is thrown in vs / on our exception tab).

niemyjski:
@ejsmith do you agree?

niemyjski:
Here is a link to the simple error model: https://github.com/exceptionless/Exceptionless.Net/blob/master/Source/Shared/Models/Client/Data/SimpleError.cs

niemyjski:
screen shot 2016-02-21 at 7 44 48 am
screen shot 2016-02-21 at 7 45 06 am
screen shot 2016-02-21 at 7 45 18 am
screen shot 2016-02-21 at 7 45 27 am

niemyjski:
@ejsmith @skoub I posted a complex event but just set the `Method.Name` value to there raw event method name. The downside to this is we don't get a break down of parameter info or generics but this is the end result.. We could write a single line of code on our clients to not display the duplicate parameters... But yeah, what are your thoughts? It kind of sucks that it makes our models be half populated and not normalized.

ejsmith:
I think that is pretty reasonable.

skoub:
What is the goal to have a Parameters property on your model? Are you doing something special with it or just showing it? Because maybe you can simplify your model?

ejsmith:
Yeah, we probably made the model too complex. We don't really get a lot of value from it being broken out so much, but I'm afraid that might be hard to change at this point. Guess we can think about it and see how hard it would be to make it simpler.

skoub:
I don't think it would be hard to merge the Parameters property to his MethodName at the parser. Later, you can make a v3 api to remove the Parameters property.

There's a few properties in your model that is .net specific, i think: IsSignatureTarget, ModuleId, GenericArguments. We will face a few problems if we want to support every languages.

niemyjski:
I don't think it hurts anything by being too broken out and I think it allows us to do some rule based things later? But I guess for now, we should do some testing with specifying the full name here and I can update the client to check for `()` and then not append them.

niemyjski:
@skoub I guess the next steps are too look through some of the clients and see if anything else is special cased.

Also, I'm wondering if raw data is always json. if it is we can try to deserialize it and then move those properties to the data object or just store the string.

skoub:
About the Parameters, i don't think there's a real value to have them separated. We just need to know what method is broken and what parameters it have. I don't need extra information. Like @ejsmith said, your model is probably going too deep to get the information.

What kind of rules can you make out of them? I'm not sure that it's usefull to search for methods that have a parameter named "username".

Keep it simple for now or you will have many hacks in your product like what we can see in the raygun family. When you will support many languages, you will have a better idea of how far you can go with the information that you collect from the client. Grow it later if there's really a need for it. Sometimes, we try to think too much ahead.

But it's your product so it's your call.

ejsmith:
@skoub I completely agree. We over baked this model. We can make it much simpler and make it much simpler to add new clients that send errors. We spent a lot of time trying to keep everything else simple and yet this model is really overly complex.

niemyjski:
Just merged in latest.

niemyjski:
- [ ] Update for batch sending events.
- [ ] Investigate places where were calling MethodExtensions.GetFullName() for things like emails and summaries. We'd need to update these use cases to just use the name property if it already contains everything.
- [ ] Check to see if raw data is a json object (it might always be). if it is, convert it to key value pair and add each item to the raw event..
- [ ] See if we should be populating the `UserInfo.Identity` to the id they specified instead of the email address they specified.
- [ ] Look over client implementations and see if they are special casing anything.

skoub:
Maybe add tasks to test each clients? At my return, we can start testing javascript, java, php, etc.

niemyjski:
I'd rather spend time on platforms we don't support first :)

niemyjski:
@skoub have you had any chance to take a look into this?

Source/Core/Plugins/EventParser/Raygun/RaygunEventMapper.cs
```diff
@@ -209,7 +212,18 @@ public class RaygunEventMapper : EventMapperBase<RaygunModel> {
...
         protected override string MapVersion(RaygunModel source) {
-            return source.Details?.Version;
+            var version = source.Details?.Version;
+
+            if (string.IsNullOrWhiteSpace(version)) {
+                return null;
+            }
+
+            // raygun send the string 'Not Supplied' when the version is not set in the client. We treat it as null instead.
+            if (version.ToLower() == "not supplied") {
```
niemyjski: This should be ToLowerInvariant() as turkish submitted events will not work :(

Source/Core/Plugins/EventParser/Raygun/RaygunEventParserPlugin.cs
```diff
@@ -0,0 +1,41 @@
...
+using Exceptionless.Core.Models;
+using Exceptionless.Core.Pipeline;
+
+namespace Exceptionless.Core.Plugins.EventParser.Raygun {
+    [Priority(0)]
+    public class RaygunEventParserPlugin : IEventParserPlugin {
+        private readonly RaygunEventMapper _mapper = new RaygunEventMapper();
+
+        public List<PersistentEvent> ParseEvents(string input, int apiVersion, string userAgent) {
+            if (userAgent != "raygun" && apiVersion != 1)
```
niemyjski: I guess we we could just do user agent != raygun  return, there api version may increase and we only want to support what we know.

Source/Core/Plugins/EventParser/Raygun/RaygunEventParserPlugin.cs
```diff
@@ -0,0 +1,41 @@
...
+                return null;
+            
+            var events = new List<PersistentEvent>();
+            switch (input.GetJsonType()) {
+                case JsonType.Object: {
```
niemyjski: What do you mean?

Source/Api/Controllers/EventController.cs
```diff
@@ -625,6 +625,15 @@ public class EventController : RepositoryApiController<IEventRepository, Persist
 
             return Ok();
         }
+        
+        [HttpPost]
+        [Route("~/entries")]
```
niemyjski: It's the end point they submit too, if we match their api, then it would automatically be handled. Would be interested in helping @skoub finish this pr? I'd be more than happy to meet up with you two and merge in the latest changes.

niemyjski: For example, you'd just need to change the server url

Source/Core/Plugins/EventParser/Raygun/RaygunEventParserPlugin.cs
nj: Codewise the following works the same: 
```cs
           switch (input.GetJsonType()) {
                case JsonType.Object:
                    RaygunModel model;
                    if (input.TryFromJson(out model)) {
                        PersistentEvent ev = _mapper.Map(model);
                        events.Add(ev);
                    }
                    
                    break;

                case JsonType.Array:
                    // ...
                    break;
            }
```

So basically it's might just be personal preference or related to readability. Haven't seen it done this way before 😄

niemyjski: Ah, I guess I never paid that close attention. I think I just automatically do it for multi line big blocks of code, it's easier to tell where it starts and ends.

Source/Core/Plugins/EventParser/Raygun/RaygunEventParserPlugin.cs
nj: Okay 😃

But better change to using *or* (`||`) instead, to fix the current issues.

`and` expectation handled:
1. **raygun** with **v1**: 👍
2. **raygun** with **v2**: 👎
3. **other** with **v1**: 👎
4. **other** with **v2**: 👍

2 + 3 will in other words, not do `return null` as expected, but passing through.

`or` will on the other hand, cover all 4 cases above.

Source/Api/Controllers/EventController.cs
nj: Haven't tried `raygun` myself - just expected one to be able to pass to a url specified by the user, like webhooks on GitHub and other service providers.

Yes, I can try help @skoub out.

niemyjski:
Here are some test cases that need to be added (https://gist.github.com/nj/462e92d1c553e8b09d899e54639e23ec)

niemyjski:
I really want to bring these changes in but due to inactivity and time constraints I'm going to close this pr.

PR #195, opened 2016-02-22
Updates to session management

niemyjski:


niemyjski:
@ejsmith do you think we should create a manual job to mark all existing heartbeat stacks as hidden?

ejsmith:
Made a couple comments, but the the changes look pretty good.

niemyjski:
Updated, now you just need to sign the cla and merge it.

PR #197, opened 2016-02-23
Updated manual stacking implementation

niemyjski:
https://github.com/exceptionless/Exceptionless.Net/pull/82

### DIFF
--- a/Source/Core/Pipeline/010_AssignToStackAction.cs
+++ b/Source/Core/Pipeline/010_AssignToStackAction.cs
@@ -58,7 +58,8 @@ namespace Exceptionless.Core.Pipeline {
                         Logger.Trace().Message("Creating new event stack.").Write();
                         ctx.IsNew = true;
 
-                        string title = _formattingPluginManager.GetStackTitle(ctx.Event);
+                        var msi = ctx.Event.GetManualStackingInfo();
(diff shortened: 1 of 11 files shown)


Source/Core/Pipeline/010_AssignToStackAction.cs
```diff
@@ -58,7 +58,8 @@ public class AssignToStackAction : EventPipelineActionBase {
                         Logger.Trace().Message("Creating new event stack.").Write();
                         ctx.IsNew = true;
 
-                        string title = _formattingPluginManager.GetStackTitle(ctx.Event);
+                        var msi = ctx.Event.GetManualStackingInfo();
+                        string title = !String.IsNullOrWhiteSpace(msi?.Title) ? msi.Title : _formattingPluginManager.GetStackTitle(ctx.Event);
```
niemyjski: Was wondering if I should make this into a stack formatting plugin... But I think this is fine.

niemyjski:
@adamzolotarev thoughts?

niemyjski:
My only thoughts is that the model and known key should be changed from StackingInfo to ManualStackingInfo thoughts?

adamzolotarev:
Looks nice, thank you!

PR #202, opened 2016-03-18
Support for improved searchbar

niemyjski:
We now automatically expand queries to include hidden and fixed terms if they are not included.


niemyjski:
@ejsmith there might be an issue with this. What if you don't care if something is marked as hidden or fixed. It will now be populated :(.

Issue #177, opened 2016-01-20
Improved notifications


niemyjski:
The goal of notifications is to keep you informed without overwhelming you with too much information. It would be nice to know when things change (marked as fixed), a project is getting above normal exceptions, etc. Below are some issues that have been raised by others on how we can make notifications better.

### Objectives
* [ ] [Additional information on all emails (tags, descriptions, user identity)](https://github.com/exceptionless/Exceptionless/issues/136).
* [ ] [Ability to configure rules based on types, tags](https://github.com/exceptionless/Exceptionless/issues/76), [time of day](https://github.com/exceptionless/Exceptionless/issues/92), or [number of exceptions](https://github.com/exceptionless/Exceptionless/issues/176).
* [ ] [Send a periodic digest email based on activity](https://github.com/exceptionless/Exceptionless/issues/69)
* [ ] [Ability to snooze notifications for a period of time](https://github.com/exceptionless/Exceptionless/issues/92).
* [ ] [Don't send as many emails if there hasn't been any changes in activity](https://github.com/exceptionless/Exceptionless/issues/176). Example a user hasn't configured there project shouldn't get a daily notification saying they are Exceptionless every day.

### Third Party Integrations
Provide a easy way to manage third party integrations. Currently these integrations can be done with zapier. **We should define what the default message should look like.**
* [ ] [Slack](https://github.com/exceptionless/Exceptionless/issues/87)
* [ ] [Hipchat](https://github.com/exceptionless/Exceptionless/issues/81)

### Suggestions
* @webcoda [suggested](https://github.com/exceptionless/Exceptionless/issues/76) that it would be really nice to be able to send an alert to any user (even if they are not part of the system).

HenrikHoyer:
Please add an "daily/weekly/monthly" option

HenrikHoyer:
Please remove the "congrats, no event was logged" mails - they are just noise

niemyjski:
@ejsmith what do you think about not sending dailies if there are no events? Could lead me to think that notifications are not working or we have not yet configured our project. We do have project is configured flag. Maybe we add configuration required emails and option to not send.

HenrikHoyer:
I want to receive exceptions from you, not the normal case :-) All monitoring systems we use only send notifications on exceptions/deviations - none are sending "everything ok mails".

But if your current users are depending on these "everything ok" mails, then add a "send summary mails even if no exceptions has been logged" flag (imho with default = no)

[...]

PR #204, opened 2016-03-23
Added the ability to send organization notifications

niemyjski:
https://github.com/exceptionless/Exceptionless/issues/177


Source/Core/Jobs/WorkItemHandlers/OrganizationNotificationWorkItemHandler.cs
```diff
@@ -0,0 +1,71 @@
...
+        public override async Task HandleItemAsync(WorkItemContext context) {
+            var workItem = context.GetData<OrganizationNotificationWorkItem>();
+            _logger.Info("Received organization notification work item for: {0} IsOverHourlyLimit: {1} IsOverMonthlyLimit: {2}", workItem.OrganizationId, workItem.IsOverHourlyLimit, workItem.IsOverMonthlyLimit);
+
+            var organization = await _organizationRepository.GetByIdAsync(workItem.OrganizationId, true).AnyContext();
+            if (organization == null)
+                return;
+            
+            if (workItem.IsOverMonthlyLimit)
+                await SendOverageNotificationsAsync(organization, workItem.IsOverHourlyLimit, workItem.IsOverMonthlyLimit).AnyContext();
```
niemyjski: I'm only sending these when you go over your monthly usage for now. So I don't think we need to rate limit it because it should only happen once via overage handler

Source/Core/Repositories/OrganizationRepository.cs
```diff
@@ -223,6 +218,11 @@ public class OrganizationRepository : RepositoryBase<Organization>, IOrganizatio
                 await Cache.SetAsync(GetUsageSavedCacheKey(organizationId), DateTime.UtcNow, TimeSpan.FromDays(32)).AnyContext();
             }
 
+            if (justWentOverMonthly)
+                await PublishMessageAsync(new PlanOverage { OrganizationId = org.Id }).AnyContext();
+            else if (justWentOverHourly)
+                await PublishMessageAsync(new PlanOverage { OrganizationId = org.Id, IsHourly = true }).AnyContext();
+
```
niemyjski: I moved this down to after the organization is saved. This ensures that when the message is published and the organization is retrieved the stats are up-to-date. So we could give detailed and up-to-date metrics when you go over your limit.

PR #227, opened 2016-04-28
Added new heartbeat api endpoint

niemyjski:
This adds new api fire and forget end point for sending heartbeats.

@ejsmith do you think we should decrease the job time and inactive session period from 10 minutes to something smaller so things run more often?

### DIFF
--- a/Source/Api/Controllers/EventController.cs
+++ b/Source/Api/Controllers/EventController.cs
@@ -449,7 +449,7 @@ namespace Exceptionless.Api.Controllers {
         /// <param name="version">The api version that should be used.</param>
         /// <param name="id">The session id or user id.</param>
         /// <param name="close">If true, the session will be closed.</param>
-        /// <response code="202">Accepted</response>
+        /// <response code="200">OK</response>
         /// <response code="400">No project id specified and no default project was found.</response>
         /// <response code="404">No project was found.</response>
         [HttpGet]
@@ -500,7 +500,7 @@ namespace Exceptionless.Api.Controllers {
         /// <param name="type">The event type</param>
         /// <param name="userAgent">The user agent that submitted the event.</param>
         /// <param name="parameters">Parameters that control what properties are set on the event</param>
-        /// <response code="202">Accepted</response>
+        /// <response code="200">OK</response>
         /// <response code="400">No project id specified and no default project was found.</response>
         /// <response code="404">No project was found.</response>
         [HttpGet]
--- a/Source/Api/Controllers/EventController.cs
+++ b/Source/Api/Controllers/EventController.cs
@@ -447,17 +447,18 @@ namespace Exceptionless.Api.Controllers {
         /// </summary>
         /// <param name="projectId">The identifier of the project.</param>
         /// <param name="version">The api version that should be used.</param>
-        /// <param name="sessionIdOrUserId">The session id or user id.</param>
+        /// <param name="id">The session id or user id.</param>
+        /// <param name="close">If true, the session will be closed.</param>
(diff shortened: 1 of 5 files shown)


Source/Core/Jobs/CloseInactiveSessionsJob.cs
```diff
@@ -36,18 +39,24 @@ public class CloseInactiveSessionsJob : JobWithLockBase {
...
                     sessionsToUpdate.Add(sessionStart);
-
                     Debug.Assert(sessionStart.Value != null && sessionStart.Value >= 0, "Session start value cannot be a negative number.");
                 }
 
-                if (sessionsToUpdate.Count > 0)
+                if (sessionsToUpdate.Count > 0) {
                     await _eventRepository.SaveAsync(sessionsToUpdate).AnyContext();
-                
+                    await _cacheClient.RemoveAllAsync(sessionsToUpdate.SelectMany(GetSessionHeartbeatCacheKeys)).AnyContext();
```
niemyjski: This is the only real issue I see. The cache could be updated with a new timestamp and we are throwing it away. The window of time this happens is very very short..

Source/Api/Controllers/EventController.cs
```diff
@@ -439,6 +443,41 @@ public class EventController : RepositoryApiController<IEventRepository, Persist
...
+        /// <param name="sessionIdOrUserId">The session id or user id.</param>
+        /// <response code="202">Accepted</response>
+        /// <response code="400">No project id specified and no default project was found.</response>
+        /// <response code="404">No project was found.</response>
+        [HttpGet]
+        [Route("~/api/v{version:int=2}/events/session/{sessionIdOrUserId:minlength(1)}/heartbeat")]
+        [Route("~/api/v{version:int=2}/projects/{projectId:objectid}/events/session/{sessionIdOrUserId:minlength(1)}/heartbeat")]
+        [OverrideAuthorization]
+        [Authorize(Roles = AuthorizationRoles.Client)]
+        public async Task<IHttpActionResult> GetSubmitHeartbeat(string projectId = null, int version = 2, string sessionIdOrUserId = null) {
```
ejsmith: I think a better name would be SendHeartbeat or RecordHeartbeat.  Get seems wrong because it's not getting anything. But we are making it a GET to just keep the implementation simple. Doesn't mean the method name has to be Get though.

Source/Api/Controllers/EventController.cs
```diff
@@ -439,6 +443,41 @@ public class EventController : RepositoryApiController<IEventRepository, Persist
...
+
+            // must have a project id
+            if (String.IsNullOrEmpty(projectId))
+                return BadRequest("No project id specified and no default project was found.");
+
+            var project = await GetProjectAsync(projectId);
+            if (project == null)
+                return NotFound();
+            
+            string decodedId = Uri.UnescapeDataString(sessionIdOrUserId).Trim();
```
ejsmith: Doesn't this happen automatically with the WebApi parameter binding?

niemyjski: I've seen cases where it doesn't, and I'm specifically going to have our client call Uri.Encode. This may need more testing but we had to do it on our check-name routes otherwise they wen't decoded.

ejsmith: I think you should verify that and not do extra work if we don't have to.

niemyjski: We don't need it.

ejsmith:
[selfie-3](https://github.com/thieman/github-selfies/)

PR #231, opened 2016-05-13
First pass at AD auth

laughinggoose:
If this is approved, I need to 

- [ ] Update self-host wiki with instructions to setup AD auth

### DIFF
--- a/Source/Api/Controllers/AuthController.cs
+++ b/Source/Api/Controllers/AuthController.cs
@@ -8,6 +8,7 @@ using Exceptionless.Api.Extensions;
 using Exceptionless.Api.Models;
 using Exceptionless.Api.Utility;
 using Exceptionless.Core;
+using Exceptionless.Core.Authentication;
 using Exceptionless.Core.Authorization;
 using Exceptionless.Core.Extensions;
 using Exceptionless.Core.Mail;
@@ -27,7 +28,8 @@ using OAuth2.Models;
 namespace Exceptionless.Api.Controllers {
     [RoutePrefix(API_PREFIX + "/auth")]
     public class AuthController : ExceptionlessApiController {
-        private readonly IOrganizationRepository _organizationRepository;
+	    private readonly IDomainLoginProvider _domainLoginProvider;
+	    private readonly IOrganizationRepository _organizationRepository;
         private readonly IUserRepository _userRepository;
         private readonly ITokenRepository _tokenRepository;
         private readonly ICacheClient _cacheClient;
@@ -36,8 +38,9 @@ namespace Exceptionless.Api.Controllers {
 
         private static bool _isFirstUserChecked;
 
-        public AuthController(IOrganizationRepository organizationRepository, IUserRepository userRepository, ITokenRepository tokenRepository, ICacheClient cacheClient, IMailer mailer, ILogger<AuthController> logger) {
-            _organizationRepository = organizationRepository;
+        public AuthController(IOrganizationRepository organizationRepository, IUserRepository userRepository, ITokenRepository tokenRepository, ICacheClient cacheClient, IMailer mailer, ILogger<AuthController> logger, IDomainLoginProvider domainLoginProvider) {
+	        _domainLoginProvider = domainLoginProvider;
+	        _organizationRepository = organizationRepository;
             _userRepository = userRepository;
             _tokenRepository = tokenRepository;
             _cacheClient = new ScopedCacheClient(cacheClient, "auth");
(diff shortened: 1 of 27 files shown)


laughinggoose:
I just realized that I'm doing a search for an email address using a username - I wonder if I could do the reverse in order to authenticate? (Use email to lookup username and then use username + password for authentication) That would prevent the need to change the UI project. I'll look into it some more.

Source/Core/Settings.cs
```diff
@@ -56,6 +56,10 @@ public class Settings : SettingsBase<Settings> {
 
         public bool EnableElasticsearchTracing { get; private set; }
 
+		public string LdapConnectionString { get; private set; }
+
+		public bool EnableActiveDirectoryAuth { get; internal set; }
```
niemyjski: I think this should be like some of our other enabled properties and just return true of the ldap connection string is set.

Source/Api/Controllers/AuthController.cs
```diff
@@ -95,33 +98,45 @@ public class AuthController : ExceptionlessApiController {
             }
 
             User user;
-            try {
-                user = await _userRepository.GetByEmailAddressAsync(model.Email);
-            } catch (Exception ex) {
-                _logger.Error().Exception(ex).Critical().Message("Login failed for \"{0}\": {1}", model.Email, ex.Message).Tag("Login").Identity(model.Email).SetActionContext(ActionContext).Write();
-                return Unauthorized();
-            }
+	        if (Settings.Current.EnableActiveDirectoryAuth && !IsAdLoginValid(model.Email, model.Password)) {
```
niemyjski: I'd move this check to after the user not null check, and then just return the token logic if it's valid otherwise return unauthorized.

Also on a side note, could you install the https://visualstudiogallery.msdn.microsoft.com/410e9b9f-65f3-4495-b68e-15567e543c58 extension, I noticed spacing and bracing styles are a different and we are trying to stay with k&r style and this plugin will help automatically apply our styles :)

Source/Api/Controllers/AuthController.cs
```diff
@@ -95,33 +98,45 @@ public class AuthController : ExceptionlessApiController {
...
             if (!user.IsActive) {
                 _logger.Error().Message("Login failed for \"{0}\": The user is inactive.", user.EmailAddress).Tag("Login").Identity(user.EmailAddress).Property("User", user).SetActionContext(ActionContext).Write();
                 return Unauthorized();
             }
 
-            if (String.IsNullOrEmpty(user.Salt)) {
-                _logger.Error().Message("Login failed for \"{0}\": The user has no salt defined.", user.EmailAddress).Tag("Login").Identity(user.EmailAddress).Property("User", user).SetActionContext(ActionContext).Write();
-                return Unauthorized();
-            }
+	        if (!Settings.Current.EnableActiveDirectoryAuth) {
```
niemyjski: If we do the check proposed above and validate the login after checking for the null user and if they are active check, then we can just return the token and not modify any of this code.

niemyjski:
I just made a few comments, I also just thought of something. What should we do on our change password and password reset controller methods? Might be easiest to not touch them, but the user would have to know that they are changing there local exceptionless password that's never applied.

laughinggoose:
A password isn't being set during account creation, so I don't think the password change works right now. Personally, I would leave it like that.

Source/Api/Controllers/AuthController.cs
```diff
@@ -94,34 +97,44 @@ public class AuthController : ExceptionlessApiController {
...
             }
 
-            User user;
-            try {
-                user = await _userRepository.GetByEmailAddressAsync(model.Email);
-            } catch (Exception ex) {
-                _logger.Error().Exception(ex).Critical().Message("Login failed for \"{0}\": {1}", model.Email, ex.Message).Tag("Login").Identity(model.Email).SetActionContext(ActionContext).Write();
-                return Unauthorized();
-            }
+	        User user;
```
niemyjski: I'm not sure why all this changed it's the same exact code, maybe due to spacing and styling

Source/Api/Controllers/AuthController.cs
```diff
@@ -183,20 +196,27 @@ public class AuthController : ExceptionlessApiController {
                 }
             }
 
+	        if (Settings.Current.EnableActiveDirectoryAuth && !IsAdLoginValid(model.Email, model.Password)) {
```
niemyjski: This check is no longer required as your doing it on line 133.

Source/Api/Controllers/AuthController.cs
```diff
@@ -183,20 +196,27 @@ public class AuthController : ExceptionlessApiController {
...
+				_logger.Error().Message("Signup failed (bad AD login) for \"{0}\"", model.Email).Tag("Signup").Identity(model.Email).SetActionContext(ActionContext).Write();
+		        return BadRequest("Username and password combination failed to authenticate with Active Directory.");
```
niemyjski: I'd just return bad request with no message, otherwise you are leaking more info to a potential attacker

Source/Api/Controllers/AuthController.cs
```diff
@@ -183,20 +196,27 @@ public class AuthController : ExceptionlessApiController {
...
             };
             user.CreateVerifyEmailAddressToken();
             user.Roles.Add(AuthorizationRoles.Client);
             user.Roles.Add(AuthorizationRoles.User);
             await AddGlobalAdminRoleIfFirstUserAsync(user);
 
-            user.Salt = Core.Extensions.StringExtensions.GetRandomString(16);
-            user.Password = model.Password.ToSaltedHash(user.Salt);
-
+	        if (!Settings.Current.EnableActiveDirectoryAuth) {
```
niemyjski: Any reason for special case it, what if you wanted to remove your ad functionality in the future, you could just login with your ad account (granted the password hasn't change) without doing a password reset.

niemyjski:
I added a few comments, other than those comments everything looks good :) e

niemyjski:
I guess lets leave it as it is for now and you shouldn't try to change your user account. I'd like to morph this code into adfs in the future so we could have generic ad logins in production against your own ad server :).

Source/Api/Controllers/AuthController.cs
laughinggoose: Is there a way to automatically fix the spacing? I tried ctrl-k-d awhile ago and it seemed to affect more than just the lines I had touch.

Source/Api/Controllers/AuthController.cs
laughinggoose: True, but AD passwords can change over time (some reset every 30 days) and there wouldn't be a way to sync them back to Exceptionless. You could have a case where someone has to remember the password they used X password resets ago in order to change their password or login after switching off AD.

Source/Api/Controllers/AuthController.cs
laughinggoose: Wouldn't this be the same as a login page? I'd hate to just say "sorry" to someone without giving them more information on why they couldn't sign up.

laughinggoose:
I've added responses to your comments. Is there anything else that needs done?

niemyjski:
Thank you for your commit, can you please update the wiki. I'm going to make a few tweaks to setting where active directory will be enabled if the connection string is set.

laughinggoose:
I've updated the wiki: https://github.com/exceptionless/Exceptionless/wiki/Self-Hosting

Issue #136, opened 2015-08-27
Improved Notification Emails


kaiukdevelopment:
Hello,

Could you please add extra information to the emails that get sent out as notifications for each exception? We would like to see:

All tags
User identity
User description

Thank You,

Robert Edworthy
KAI Consulting

niemyjski:
It would be good to be able to click on the email address and prepopulate the subject with the event's message.

ejsmith:
I would love to see more information on these emails as well. @kaiukdevelopment, would you be willing to make these changes? I believe the we currently pass in the entire event object and it would just be a matter of changing the html template.

niemyjski:
@kaiukdevelopment would you mind submitting a pull request for this?

niemyjski:
I'm closing this as it's now merged into 177.

niemyjski:
@kaiukdevelopment I wanted to let you know that this is currently being implemented in #177 and rolled soon :)

PR #235, opened 2016-06-10
Added user information and tags to notice email

laughinggoose:
For #177 

This pull request is incomplete. I wanted to get feedback on the new plain text email before I copied the changes to the HTML email.

Here is the plain text output with user info:

> A new  critical  event has occurred in the "Testing" project.
> 
> Tags:
> important, web
> 
> 
> User's Name:
> First Last
> 
> 
> User's Identity:
> firstylaster
> 
> 
> User's Email:
> useremail@domain.com
> 
> 
> User's Description:
> A person
> 
> 
> View the event:
> http://localhost:50000/event/1

### DIFF
--- a/Source/Core/Mail/Templates/Notice/PlainText.cshtml
+++ b/Source/Core/Mail/Templates/Notice/PlainText.cshtml
@@ -1,5 +1,8 @@
 ﻿@using System
+@using System.Linq
+@using Exceptionless
 @using Exceptionless.Core.Extensions
+@using Exceptionless.Core.Models.Data
 @inherits RazorSharpEmail.EmailTemplate<Exceptionless.Core.Mail.Models.EventNotificationModel>
 @if (Model.IsNew) {
 <text>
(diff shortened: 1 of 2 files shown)


niemyjski:
Thanks for this, We'll take a look at this and get it merged soon.

ejsmith:
I like it! I am thinking we should collapse the user information down into a single line though. It's probably pushing important content down out of view.

Not sure if we need the users description. If we do want it, we can hide that section if their description is empty.

laughinggoose:
@ejsmith How would you condense the user information into a single line? What if one of the user fields are missing?

The user's description was mentioned specifically in #136. I'm not sure what is though really. I don't think we set this for our users.

ejsmith:
So for user we just have identity and name:
https://github.com/exceptionless/Exceptionless.Net/blob/master/src/Exceptionless/Models/Client/Data/UserInfo.cs

The user description is where the user can type in a description of what happened for an error after the error occurs. So I think we can include that if it is populated which is not very often. So maybe something like this:

Tags:
important, web

User:
name (identity)

User's Description:
I was trying to place an order.

niemyjski:
@laughinggoose Is this pr complete?

laughinggoose:
Not yet. I still need to update the text email with Eric's feedback and then port it to the HTML email.

niemyjski:
@laughinggoose I'm going to close this until it's ready. Can you please reopen a pr when it's finished.

Issue #239, opened 2016-07-15
Bug in resolving EnableActiveDirectoryAuth


crowbarsolutions:
So I'm testing out the latest release, trying to get active directory authentication working. I followed the wiki and set the LdapConnectionString setting in my web.config. However AD authentication did not work.

After poking around your code it looks like when the application calls `Settings.Current.EnableActiveDirectoryAuth` in AuthController.cs it still returns false even if the LdapConnectionString setting is set.

I managed to work around this by simply adding the EnableActiveDirectoryAuth setting into my web.config and setting it to "true".

My guess is there is a bug in how the `GetBool()` method is being called [in this code](https://github.com/exceptionless/Exceptionless/blob/f3cb33c8f6192f1449070c1ecc5facd3646d5e81/Source/Core/Settings.cs#L179)

I've updated the wiki to save people from running into this bug.

niemyjski:
@laughinggoose Any chance you can submit a pr for setting the Enable setting to true if the connection string is set there instead.

crowbarsolutions:
So the issue actually was partially my fault, I specified the LDAP connectionstring in the `<appSettings>` section of the web.config. After looking at the code it became clear the application was looking for it in the `<connectionStrings>` section of the web.config file. I just went in and made the Wiki instructions much less ambiguous.

I also submitted #240 which includes an empty `LdapConnectionString` setting in the web.config to act as a template, which is more consistent with how the Redis and ElasticSearch connection strings are handled. Sorry about the NL at the EOF, github seems to add that to the end no matter what I do.

niemyjski:
It's all good thanks for the heads up :)

PR #240, opened 2016-07-15
Added empty LdapConnectionString to web.config

crowbarsolutions:
This should reduce the incidence of issue #239 and make it more consistent with how other connection string are handled in the application.


niemyjski:
Can you also change this line https://github.com/exceptionless/Exceptionless/blob/f3cb33c8f6192f1449070c1ecc5facd3646d5e81/Source/Core/Settings.cs#L179 to just set the value of the result !String.IsNullOrEmpty

niemyjski:
Would you mind signing the cla and I'll merge the connecting string change in.

crowbarsolutions:
Signed it.

niemyjski:
Thanks

PR #244, opened 2016-08-01
Fixed an issue where an unsuspended organization couldn't submit events for an hour.

niemyjski:


Source/Core/Jobs/EventPostsJob.cs
```diff
@@ -89,9 +89,10 @@ public class EventPostsJob : QueueJobBase<EventPost> {
 
                 // Discard any events over there limit.
                 events = events.Take(eventsToProcess).ToList();
-
-                // Increment by count - 1 since we already incremented it by 1 in the OverageHandler.
-                await _organizationRepository.IncrementUsageAsync(project.OrganizationId, false, events.Count - 1, applyHourlyLimit: false).AnyContext();
+                
+                // Increment the count if greater than 1, since we already incremented it by 1 in the OverageHandler.
+                if (events.Count > 1)
+                    await _organizationRepository.IncrementUsageAsync(project.OrganizationId, false, events.Count - 1, applyHourlyLimit: false).AnyContext();
```
niemyjski: @ejsmith I'm wondering if we should just be incrementing the usage by the total number of events - 1 before we do the take. Otherwise we won't know the true number of events submitted.

Source/Core/Repositories/OrganizationRepository.cs
```diff
@@ -234,6 +226,21 @@ public class OrganizationRepository : RepositoryBase<Organization>, IOrganizatio
...
+
+            int monthlyEventLimit = org.GetMaxEventsPerMonthWithBonus();
+            if ((monthlyTotal - monthlyBlocked) > monthlyEventLimit)
+                return monthlyTotal - monthlyBlocked - monthlyEventLimit;
+
+            int hourlyEventLimit = org.GetHourlyEventLimit();
+            if (applyHourlyLimit && (hourlyTotal - hourlyBlocked) > hourlyEventLimit)
+                return hourlyTotal - hourlyBlocked - hourlyEventLimit;
+
+            return count;
```
niemyjski: I greatly simplified the logic from above. I think we were over thinking it.. The current logic passes all tests. Please give it a look and let me know.

Source/Core/Jobs/EventPostsJob.cs
niemyjski: @ejsmith ^^

ejsmith: Well the ones that aren't taken should get added to the blocked count, right?

niemyjski: Yeah, but that's not what we are currently doing. When we call the increment overage we just see a boolean if anything was blocked, but not how many.

PR #248, opened 2016-09-09
Updates to the new Foundatio.Repositories

niemyjski:
- Many bug fixes around querying, paging, deleting and bulk operations
- Many improvements to document indexing (Ensure document versions are synced, Ensure index is created and yellow before indexing) and many other improvements
- Easy upgrade path to Elasticsearch 2.x+


Source/Api/Controllers/EventController.cs
```diff
@@ -123,39 +123,38 @@ public class EventController : RepositoryApiController<IEventRepository, Persist
...
             var skip = GetSkip(page, limit);
             if (skip > MAXIMUM_SKIP)
-                return Ok(Enumerable.Empty<PersistentEvent>());
+                return Ok(EmptyModels);
 
             var sortBy = GetSort(sort);
             var options = new PagingOptions { Page = page, Limit = limit };
 
-            var pr = QueryProcessor.Process(userFilter);
+            var pr = QueryProcessor.Process(filter);
```
ejsmith: We need to work on getting rid of this, but doesn't have to be done right away. It should be implemented as settings on the query parser for the elastic type / repo.

niemyjski: If the query is bad, we want to return immediately with the error. Not run an invalid query and return unexpected results.

Libraries/elasticsearch.yml
```diff
@@ -1,25 +1,15 @@
...
-http.cors.allow-origin: http://dashboards.localhost.com
 
-#cloud:
-#    azure:
-#        storage:
-#            account: your_azure_storage_account
-#            key: your_azure_storage_key
+# non-ssd configuration
+#index.merge.scheduler.max_thread_count: 1
+#index.translog.flush_threshold_size: 1gb
```
niemyjski: @ejsmith We need to come up with a good convention for single node vs cluster configuration for our yml files / docs.

Issue #228, opened 2016-05-09
Problem with Turkish culture


muhammetsahin:
Hi,

I want to setup exceptionless on my machine. My machine culture is Turkish. And i got error with new project 

`{
   "took": 8,
   "timed_out": false,
   "_shards": {
      "total": 10,
      "successful": 10,
      "failed": 0
   },
   "hits": {
      "total": 5,
      "max_score": 1,
      "hits": [
         {
            "_index": "organizations",
            "_type": "project",
            "_id": "5730b170a714ca3c24feb362",
            "_score": 1,
            "_source": {
               "ıd": "5730b170a714ca3c24feb362",
               "organization_ıd": "5730b16fa714ca3c24feb361",
               "name": "Bayinet",
               "ıs_configured": false,
               "configuration": {
                  "version": 0,
                  "settings": {
                     "@@UserAgentBotPatterns": "*bot*,*crawler*,*spider*,*aolbuild*,*teoma*,*yahoo*"
                  }
               },
               "notification_settings": {
                  "5730b15ba714ca3c24feb360": {
                     "send_daily_summary": true,
                     "report_new_errors": true,
                     "report_critical_errors": true,
                     "report_event_regressions": true,
                     "report_new_events": false,
                     "report_critical_events": false
                  }
               },
               "delete_bot_data_enabled": true,
               "next_summary_end_of_day_ticks": 635984388000000000,
               "created_utc": "2016-05-09T15:49:04.1535252Z",
               "modified_utc": "2016-05-09T15:49:04.1535252Z"
            }
         },
         {
            "_index": "organizations",
            "_type": "organization",
            "_id": "5730b16fa714ca3c24feb361",
            "_score": 1,
            "_source": {
               "ıd": "5730b16fa714ca3c24feb361",
               "name": "Penta",
               "plan_ıd": "EX_UNLIMITED",
               "plan_name": "Unlimited",
               "plan_description": "Unlimited",
               "billing_change_date": "2016-05-09T18:49:03.8720377+03:00",
               "billing_changed_by_user_ıd": "5730b15ba714ca3c24feb360",
               "billing_status": 0,
               "billing_price": 0,
               "max_events_per_month": -1,
               "bonus_events_per_month": 0,
               "retention_days": -1,
               "ıs_suspended": false,
               "has_premium_features": true,
               "max_users": -1,
               "max_projects": -1,
               "created_utc": "2016-05-09T15:49:03.8720377Z",
[...]

PR #261, opened 2016-12-30
Turkish culture problem solved.

barankaynak:
#228 
ToLower methods changed with ToLowerInvariant method.


niemyjski:
Thank you for the pull request! I'll try and put together a minor release here in the next few days.

PR #265, opened 2017-01-18
Suppress auto-reply

nj:
Any receiver having auto-reply enabled will result in an auto-reply mail being sent back to the sender address for any notifications sent out from Exceptionless. I.e. for no-reply addresses will end up bouncing with several retries and mails received as a consequence.

By including certain headers to the mails sent will make the receiver (server) skip sending auto-reply for these types of mails.

Details of the related headers:
- http://www.redmine.org/issues/19558
- https://www.jitbit.com/maxblog/18-detecting-outlook-autoreplyout-of-office-emails-and-x-auto-response-suppress-header/


niemyjski:
Thanks for this! Are you self hosting by any chance?

nj:
@niemyjski Yes 😃 Have used the hosted solution previously though.

nj:
@niemyjski Thanks for the quick merge :wink:

Hope it will become handy for others as well.

ejsmith:
Thanks!

PR #273, opened 2017-02-07
Ability to quickly discard large compressed events

niemyjski:


src/Exceptionless.Core/Jobs/EventPostsJob.cs
```diff
@@ -30,47 +30,77 @@ public class EventPostsJob : QueueJobBase<EventPost> {
...
+            _jsonSerializerSettings = jsonSerializerSettings;
 
             AutoComplete = false;
         }
 
         protected override async Task<JobResult> ProcessQueueEntryAsync(QueueEntryContext<EventPost> context) {
             var queueEntry = context.QueueEntry;
-            var eventPostInfo = await _storage.GetEventPostAndSetActiveAsync(queueEntry.Value.FilePath, _logger, context.CancellationToken).AnyContext();
-            if (eventPostInfo == null) {
+            var fileInfo = await _storage.GetFileInfoAsync(queueEntry.Value.FilePath).AnyContext();
```
niemyjski: I need to add a test to Foundatio to see what happens if this path is null or doesn't exist

src/Exceptionless.Core/Jobs/EventPostsJob.cs
```diff
@@ -30,47 +30,77 @@ public class EventPostsJob : QueueJobBase<EventPost> {
...
+            if (fileInfo == null) {
+                await queueEntry.AbandonAsync().AnyContext();
```
niemyjski: If the result is null lets bail quick

src/Exceptionless.Core/Jobs/EventPostsJob.cs
```diff
@@ -30,47 +30,77 @@ public class EventPostsJob : QueueJobBase<EventPost> {
...
+                return JobResult.FailedWithMessage($"Unable to retrieve post data info '{queueEntry.Value.FilePath}'.");
+            }
+
+            if (fileInfo.Size > Settings.Current.MaximumEventPostSize) {
+                await queueEntry.CompleteAsync().AnyContext();
+                return JobResult.FailedWithMessage($"Unable to process post data '{queueEntry.Value.FilePath}' ({fileInfo.Size} bytes): Maximum event post size limit ({Settings.Current.MaximumEventPostSize} bytes) reached.");
```
niemyjski: If it's greater than max size lets not even process it.

src/Exceptionless.Core/Jobs/EventPostsJob.cs
```diff
@@ -30,47 +30,77 @@ public class EventPostsJob : QueueJobBase<EventPost> {
...
+            }
+
+            var ep = await _storage.GetEventPostAndSetActiveAsync(queueEntry.Value.FilePath, _logger, context.CancellationToken).AnyContext();
```
niemyjski: This hits storage again (we added an extra hit just to get metadata above)

src/Exceptionless.Core/Jobs/EventPostsJob.cs
```diff
@@ -30,47 +30,77 @@ public class EventPostsJob : QueueJobBase<EventPost> {
...
+                await CompleteEntryAsync(queueEntry, ep, SystemClock.UtcNow).AnyContext();
                 return JobResult.Success;
             }
 
+            byte[] uncompressedData = ep.Data;
+            if (!String.IsNullOrEmpty(ep.ContentEncoding)) {
+                try {
+                    uncompressedData = await uncompressedData.DecompressAsync(ep.ContentEncoding).AnyContext();
+                } catch (OutOfMemoryException ex) {
+                    await CompleteEntryAsync(queueEntry, ep, SystemClock.UtcNow).AnyContext();
```
niemyjski: Calling `uncompressedMemoryStream.ToArray()` to get the bytes can throw an out of memory exception. IF this is the case bail!

src/Exceptionless.Core/Jobs/EventPostsJob.cs
```diff
@@ -30,47 +30,77 @@ public class EventPostsJob : QueueJobBase<EventPost> {
...
+                    return JobResult.FailedWithMessage($"Unable to decompress post data '{queueEntry.Value.FilePath}' ({ep.Data.Length} bytes compressed): {ex.Message}");
+                }
+
+                long maxUncompressedEventPostSize = Settings.Current.MaximumEventPostSize * 10;
+                if (uncompressedData.Length > maxUncompressedEventPostSize) {
```
niemyjski: I added the check which is 10x the max event post size. We could still get large events through (if it's a single event it could be 10x the limit :\.

src/Exceptionless.Core/Jobs/EventPostsJob.cs
```diff
@@ -138,51 +168,43 @@ public class EventPostsJob : QueueJobBase<EventPost> {
             }
 
             foreach (var requeueEvent in eventsToRetry) {
+                string contentEncoding = null;
+                byte[] data = requeueEvent.GetBytes(_jsonSerializerSettings);
+                if (data.Length > 1000) {
+                    data = await data.CompressAsync().AnyContext();
+                    contentEncoding = "gzip";
+                }
```
niemyjski: @ejsmith I'm wondering if we should even be recompressing it right here.. If it's under our max allowed bytes we probably shouldn't even be dequeueing it.. If we don't compress it here, the check above to get metadata will instantly complete it.

src/Exceptionless.Core/Jobs/EventPostsJob.cs
```diff
@@ -30,47 +30,77 @@ public class EventPostsJob : QueueJobBase<EventPost> {
...
             var queueEntry = context.QueueEntry;
-            var eventPostInfo = await _storage.GetEventPostAndSetActiveAsync(queueEntry.Value.FilePath, _logger, context.CancellationToken).AnyContext();
-            if (eventPostInfo == null) {
+            var fileInfo = await _storage.GetFileInfoAsync(queueEntry.Value.FilePath).AnyContext();
+            if (fileInfo == null) {
+                await queueEntry.AbandonAsync().AnyContext();
+                return JobResult.FailedWithMessage($"Unable to retrieve post data info '{queueEntry.Value.FilePath}'.");
+            }
+
+            if (fileInfo.Size > Settings.Current.MaximumEventPostSize) {
```
nj: Properly a bit redundant, as basically already checked through the `OverageHandler`?

niemyjski: Yes, but it's probably to be safer than sorry, since we are getting massive events in storage. Also overage handler might not work if the content length isn't sent or is malformed??

nj: Yeah, but I guess if ContentLength is not set or even not matching up, there's much worse issues at stake.

PR #278, opened 2017-02-16
Default BaseURL is misleading

caesay:
If URL re-writing is not enabled, the base url in the app config must end with "/#!" or links in emails will not work. (verification email link ends up redirecting to the dashboard, and doesn't actually verify the email)


niemyjski:
Thanks!!!

PR #291, opened 2017-03-31
Add support for MailKit (290)

### LINKED ISSUE #290: Cannot send email using the 465 smtp port.
According to the document of microsoft, the System.Net.Mail cannot send email message through 465 port(https://blogs.msdn.microsoft.com/webdav_101/2008/06/02/system-net-mail-with-ssl-to-authenticate-against-port-465/).
But we want to use the tencent exmail account which have to use 465 port, so the mail sender component should be replaced another third part solution instead of System.Net.Mail.

caesay:
Hi, 
I'm having trouble finding the latest Foundatio nuget packages, and C#7 is broken for me right now - so I can't build and this is un-tested. Feel free to close/disregard it if you're looking for something a little more production ready.

Closes #290


ejsmith:
Thanks @caesay! I think this is a good change for us. We will take a look and get it merged in.

src/Exceptionless.Core/Mail/SmtpMailSender.cs
```diff
@@ -20,16 +21,27 @@ public class SmtpMailSender : IMailSender {
...
             message.Headers.Add("Auto-Submitted", "auto-generated");
 
-            var client = new SmtpClient();
-            if (!String.IsNullOrEmpty(Settings.Current.SmtpHost)) {
-                client.Host = Settings.Current.SmtpHost;
-                client.Port = Settings.Current.SmtpPort;
-                client.EnableSsl = Settings.Current.SmtpEnableSsl;
-                client.Credentials = new NetworkCredential(Settings.Current.SmtpUser, Settings.Current.SmtpPassword);
-            }
+            using (var client = new SmtpClient()) {
```
niemyjski: Since this class is a singleton (via DI) is it safe to reuse this client and just create it once and connect once?

src/Exceptionless.Core/Extensions/MailerExtensions.cs
```diff
@@ -29,7 +30,27 @@ public static class MailerExtensions {
             return notification;
         }
 
-        public static System.Net.Mail.MailMessage ToMailMessage(this MailMessage notification) {
+        public static MimeMessage ToMailMessage(this MailMessage notification) {
+            var message = new MimeMessage();
+            var builder = new BodyBuilder();
+
+            if (!String.IsNullOrEmpty(notification.From))
```
niemyjski: I found a bug yesterday and fixe in the old implementation, where the web.config settings can't be picked up from azure functions (and has been removed now..) We need to add an else clause here kinda like this to fix this issue:

```
            if (!String.IsNullOrEmpty(notification.From))
                message.From = new MailAddress(notification.From);
            else
                message.From = new MailAddress(Settings.Current.SmtpFrom);
```

niemyjski:
Thanks for the PR, I left a few comments, I'm going to merge this today. I was thinking about moving all of this code into the insulation project so we don't have all these dependencies in the core project (I can make that change, unless you want to take a stab at it).

src/Exceptionless.Core/Mail/SmtpMailSender.cs
caesay: Yes, I can't find any official remarks on this, my initial thoughts would've been that having a long running smtp connection for unobservable benefits is not worth the complexity.

I was unable to find anything in the [MailKit source](https://github.com/jstedfast/MailKit/blob/e8338c43b44d013b5384b16bbc46f532bccd8682/MailKit/Net/Smtp/SmtpClient.cs#L1660) that indicates it will try to re-connect if it becomes disconnected, so we would have to handle the case where it throws and re-connect manually.

caesay:
I've made the change regarding the smtp from field. 

I could take a stab at moving this to Insulation (not today, probably) but what would the ideal plan be for the extension methods? Should we provide an IMailMessageConverter interface or something that can also be in Insulation?

It would help if I could build the project: I still have no idea where to find the prerelease Foundatio packages as they don't seem to be resolving through nuget.org for me. :)

caesay:
Also fwiw, the `InternetAddressList.Parse` method (as seen in the other pull request) would be required for converting a list of addresses (in a string) to an array of `MimeKit` addresses. You can [read the remarks here](https://github.com/jstedfast/MimeKit/blob/master/MimeKit/InternetAddressList.cs), but the summary is that an `InternetAddress` can be either a:
- `GroupAddress` (can contain a list of both `MailboxAddress` and/or nested `GroupAddress`)
- `MailboxAddress` (is a single mailbox address)

I'm not sure what the physical string representation of the above is, but I'm pretty sure we don't use it yet.

niemyjski:
We support groups for development or we overwrite the to field. I need to take another look at that. I'll also look at the usages, if the to field is ever multiple then it makes sense to do the parse. I'm a bit back logged at the minute but I want to merge this in real soon. Thanks again!

niemyjski:
I'm going to merge this into an email branch and move the code into the insulation project and then do some testing. If you find any issues, please let me know by commenting on this pr or submitting a pr for my  branch! Thanks again!!!!!!!!!!!!!!! Also, do you need a ci build with these changes or are you good to go as is?

PR #292, opened 2017-04-02
Fix issue #290: Cannot send email using the 465 smtp port.

edwardmeng:
Implement the SmtpMailSender with MailKit instead of System.Net.Mail to enable implicit ssl with 465 port.

### DIFF
--- a/src/Exceptionless.Core/Exceptionless.Core.csproj
+++ b/src/Exceptionless.Core/Exceptionless.Core.csproj
@@ -44,6 +44,9 @@
     <Reference Include="AutoMapper, Version=6.0.2.0, Culture=neutral, PublicKeyToken=be96cd2c38ef1005, processorArchitecture=MSIL">
       <HintPath>..\..\packages\AutoMapper.6.0.2\lib\net45\AutoMapper.dll</HintPath>
     </Reference>
+    <Reference Include="BouncyCastle.Crypto, Version=1.8.1.0, Culture=neutral, PublicKeyToken=0e99375e54769942">
+      <HintPath>..\..\packages\BouncyCastle.1.8.1\lib\BouncyCastle.Crypto.dll</HintPath>
+    </Reference>
     <Reference Include="Elasticsearch.Net, Version=5.0.0.0, Culture=neutral, PublicKeyToken=96c599bbe3e70f5d, processorArchitecture=MSIL">
       <HintPath>..\..\packages\Elasticsearch.Net.5.2.0\lib\net46\Elasticsearch.Net.dll</HintPath>
       <Private>True</Private>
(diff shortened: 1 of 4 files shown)


niemyjski:
Thanks for your PR! We greatly appreciate it. It looks like we have 2 pr's for the same thing (A great problem to have). I'm going to take https://github.com/exceptionless/Exceptionless/pull/291 and incorporate your fixes into it as well. 

Did you run into an issue with (
 +                message.To.AddRange(MimeKit.InternetAddressList.Parse(notification.To));
 +            }
 +            if (!string.IsNullOrEmpty(notification.From)) {
 +                message.From.AddRange(MimeKit.InternetAddressList.Parse(notification.From));)?

having to parse these? Also, I was wondering if it would be best to make the client be a singleton (since the smtp sender is a singleton in the di) then we only need to create and connect once and it's used quite a bit. Thoughts on this?

edwardmeng:
I just want to make it compatible with multiple receptors and senders. If there is only one receptor for every email message, I think it's ok for your code.
Another hand, I have checked the source code of mimekit. The constructor of MailBoxAddress will not verify the format of email address, but the Parse method will verify.

edwardmeng:
There is another problem in the `MailKitSender.SendAsync(MailMessage model)` method. If there is any error after the smtp client connected(e.g. failed authentication or failed send message), the DisconnectAsync will not be invoked. It means only the Dispose method will be invoked, but not send the `QUIT` command to the smtp server. So I think the source code should be wrapped with try finally block as following:
`public async Task SendAsync(MailMessage model) {
	............
	using (var client = new SmtpClient()) {
		client.ServerCertificateValidationCallback = (s, c, h, e) => true;
		await client.ConnectAsync(Settings.Current.SmtpHost, Settings.Current.SmtpPort, Settings.Current.SmtpEnableSSL).AnyContext();
		try
		{
			// Note: since we don't have an OAuth2 token, disable the XOAUTH2 authentication mechanism.
			client.AuthenticationMechanisms.Remove("XOAUTH2");
			if ((Settings.Current.SmtpUser == null) != (Settings.Current.SmtpPassword == null))
				throw new ArgumentException("Must specify both the SmtpUser and the SmtpPassword, or neither.");
			if (Settings.Current.SmtpUser != null)
				await client.AuthenticateAsync(Settings.Current.SmtpUser, Settings.Current.SmtpPassword).AnyContext();
			await client.SendAsync(message).AnyContext();
			// we don't care if there is an error at this point.
			Interlocked.Increment(ref _totalSent);
		}
		finally
		{
			await client.DisconnectAsync(true).AnyContext();
		}
	}
}`

niemyjski:
@edwardmeng very nice finds! Would you be able to submit a pr from the code in the mail message branch for the parsing as well as individual checks around connect, send and disconnect and we can write out a log message (via the injected ILogger). 

Also, the using block will force dispose to be called like you stated, what will happen if you don't call DisconnectAsync? Won't the client clean itself up? I never looked at what disconnectasync did.

niemyjski:
Hmm, I guess if we add try catches there, then the job that calls this won't blow up... And if it doesn't blow up emails won't be retried. I guess we could at the least change it to return a Task<bool> and return true or false, or just keep blowing up what are your thoughts?

I think that what you have for the snippet gives the existing functionality, but what if disconnect blows up too? I guess that is fine.

edwardmeng:
I have wrapped the implementation with try finally there, so the call this will blow up as normal. It is not try catches. So the email will be retried.
The implementation of the method `SmtpClient.Disconnect` as following
`public override void Disconnect(bool quit, CancellationToken cancellationToken = new CancellationToken())
{
    this.CheckDisposed();
    if (this.IsConnected)
    {
        if (quit)
        {
            try
            {
                this.SendCommand("QUIT", cancellationToken);
            }
            catch (OperationCanceledException)
            {
            }
            catch (SmtpProtocolException)
            {
            }
            catch (SmtpCommandException)
            {
            }
            catch (IOException)
            {
            }
        }
        this.Disconnect();
    }
}
`
The implementation of the Dispose method is as following:
`protected override void Dispose(bool disposing)
{
    if (disposing && !this.disposed)
    {
        this.disposed = true;
        this.Disconnect();
    }
    base.Dispose(this.disposed);
}
`
So the difference is whether to send the `QUIT` command to smpt server, which I have mentioned in the previous comments. But I have not studied the affection that not to send this command.

niemyjski:
Just a heads up, you can do GitHub code formatting with three ` (```cs) with cs being csharp formatting :) then at the end of your code it's ``` to finish the block.

I guess it probably wouldn't hurt to keep calling it, I wonder what the system.net.mail client was doing under the covers it :)

edwardmeng:
I have learned it, thanks very much. It's great.

PR #294, opened 2017-04-05
Added support for mailkit

niemyjski:
Initially merged in from #290 and inspected #291 as well. The only changes I made was I moved the mail kit implementation to the insulation project and cleaned up the main mailer class to remove instances of `system.net.mail`.

I also believe `To` should only ever contain 1 mail address from looking at the code. I'm not sure what will happen for those who were using mail pickup via the `<mailSettings>`. This is the right step to take tho, as it brings us one step closer to core combat.

@caesay @edwardmeng were you able to test this at all?


niemyjski:
I thought I seen a reply regarding mail kit in the past week but I don't see it now. Do you remember what it was @caesay @edwardmeng? Anyone aware of any issues with this? I'm going to test this in the morning and get it merged in pending any feedback or issues.

caesay:
I don't remember seeing anything, no - but I'd recommend we use the `InternetAddressList.Parse` method from the other pull if we're not already using it in this one.

niemyjski:
This has been updated :)

build/Start-ElasticSearch.ps1
```diff
@@ -8,98 +8,127 @@
...
 Push-Location $PSScriptRoot
 
 If (-Not (Test-Path -Path "elasticsearch-$Version") -And -Not (Test-Path -Path "elasticsearch-$Version.zip")) {
-    Write-Output "Downloading Elasticsearch $Version..."
-    Invoke-WebRequest "https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-$Version.zip" -OutFile "elasticsearch-$Version.zip"
+  Write-Output "Downloading Elasticsearch $Version..."
+  Invoke-WebRequest "https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-$Version.zip" -OutFile "elasticsearch-$Version.zip"
 } Else {
-    Write-Output "Using already downloaded Kibana $Version..."
+  Write-Output "Using already downloaded Kibana $Version..."
```
niemyjski: @ejsmith looks like we messed this up in our other scripts.

niemyjski:
I did a round of testing and everything seems to work great now. I changed the default smtp port from 587 to 465.

PR #297, opened 2017-04-24
#177 - Upgraded email templates

niemyjski:
We rewrote our email templates in zurb and also added more metadata and actions to emails.


niemyjski:
LGTM

Issue #117, opened 2015-07-09
User Access


meharhay:
It would be great to restrict access to specific projects by user in an organization.  We have different teams and the global access to all projects is not ideal for security reasons.

niemyjski:
Do you see users implemented at a organization scope and at a project level?

meharhay:
I'd see it that you would have users at the organization scope to then select which projects in that organization the user has access to.

niemyjski:
Do you want me to reopen this? Yeah, we would need something at the org and project level

meharhay:
Sorry, hit the wrong button "Close and Comment" instead of just Comment.  I reopened. Thanks!

niemyjski:
I'm closing as a duplicate to this: https://github.com/exceptionless/Exceptionless/issues/152

PR #300, opened 2017-05-13
#117 Added slack support

niemyjski:
This is just the first pass and adds support for event notifications. 

*NOTE*: I still need to add the notification end points so I can update them via the ui.


src/Exceptionless.Core/Models/SlackToken.cs
```diff
@@ -0,0 +1,91 @@
...
+            [JsonProperty("fields")]
+            public List<SlackAttachmentFields> Fields { get; set; } = new List<SlackAttachmentFields>();
+            [JsonProperty("mrkdwn_in")]
+            public string[] SupportedMarkdownFields { get; set; } = { "text", "fields" };
+            [JsonProperty("ts")]
+            public long TimeStamp { get; set; }
+        }
+
+        public class SlackAttachmentFields {
+            [JsonProperty("title")]
```
niemyjski: I put all these serialization attributes on here because it's calling an external service and I just wanted to be a 100% sure we always serialize it to their format...

PR #309, opened 2017-06-20
Fix bug the previous page url and next page url will be generated incorrectly

edwardmeng:
Preparement: We have posted many logs with the same Chinese tag "测试"
Scenario: In the search box enter keyword "测试", there will be many pages records. But when clicking the next page button, the page will be just refreshed and stay at the first page instead of navigating to next page records correctly.
Root cause: The previous page url and next page url will be generated incorrectly when search with Chinese characters.
While debugging source code, the baseUrl will be like
```
https://mywebsite/api/v2/stacks/frequent?filter=fixed:false+hidden:false+(测试)&limit=10&mode=summary&offset=480m
```
and the value of url.Query will be like
```
?filter=fixed:false+hidden:false+(%E6%B5%8B%E8%AF%95)&limit=10&mode=summary&offset=480m
```
So when replace the query to empty string, nothing will happen.
I have intended to created new extension method to generate base urls to replace the original approach.

### DIFF
--- a/src/Exceptionless.Core/Extensions/UriExtensions.cs
+++ b/src/Exceptionless.Core/Extensions/UriExtensions.cs
@@ -10,7 +10,7 @@ namespace Exceptionless.Core.Extensions {
         }
 
         public static string ToQueryString(this IEnumerable<KeyValuePair<string, string>> collection) {
-            return collection.ToConcatenatedString(pair => pair.Key == null ? pair.Value : $"{pair.Key}={System.Web.HttpUtility.UrlEncode(pair.Value)}", "&");
+            return collection.ToConcatenatedString(pair => pair.Key == null ? pair.Value : $"{pair.Key}={Uri.EscapeDataString(pair.Value)}", "&");
         }
 
         /// <summary>
--- a/src/Exceptionless.Core/Extensions/UriExtensions.cs
+++ b/src/Exceptionless.Core/Extensions/UriExtensions.cs
@@ -10,7 +10,7 @@ namespace Exceptionless.Core.Extensions {
         }
 
         public static string ToQueryString(this IEnumerable<KeyValuePair<string, string>> collection) {
-            return collection.ToConcatenatedString(pair => pair.Key == null ? pair.Value : $"{pair.Key}={pair.Value}", "&");
+            return collection.ToConcatenatedString(pair => pair.Key == null ? pair.Value : $"{pair.Key}={System.Web.HttpUtility.UrlEncode(pair.Value)}", "&");
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


niemyjski:
Thanks for the pr, I'm not completely following. So your saying that the round trip it's incorrect because the replace changes the encoding?

edwardmeng:
The original source code is following
```cs
string baseUrl = url.ToString(); 
if(!String.IsNullOrEmpty(url.Query))
{
    baseUrl = baseUrl.Replace(url.Query, "");
}
```
At the first line the value of baseUrl will be
```
https://mywebsite/api/v2/stacks/frequent?filter=fixed:false+hidden:false+(测试)&limit=10&mode=summary&offset=480m
```
It has been decoded by .net framework automatically.
At the last line, the value of url.Query will be
```
?filter=fixed:false+hidden:false+(%E6%B5%8B%E8%AF%95)&limit=10&mode=summary&offset=480m
```
which will not be decoded by .net framework automatically.
In my mind, the original purpose is just to truncate the query string from the url. But in this scenario, the replace operation will do nothing, since the value of url.Query is different from the value of baseUrl for the different encoding behaviors.

edwardmeng:
Another change has been submitted to fix the encoding issue. When sending Chinese characters to client directly, the ajax response cannot recognize then correctly.

niemyjski:
Awesome find! Is this completed and ready to go?

niemyjski:
@edwardmeng, please let me know when this is ready to be merged in.

edwardmeng:
Yes, it's ready. I have tested it.

src/Exceptionless.Core/Extensions/UriExtensions.cs
```diff
@@ -10,7 +10,7 @@ public static class UriExtensions {
         }
 
         public static string ToQueryString(this IEnumerable<KeyValuePair<string, string>> collection) {
-            return collection.ToConcatenatedString(pair => pair.Key == null ? pair.Value : $"{pair.Key}={pair.Value}", "&");
+            return collection.ToConcatenatedString(pair => pair.Key == null ? pair.Value : $"{pair.Key}={System.Web.HttpUtility.UrlEncode(pair.Value)}", "&");
```
niemyjski: I didn't see this in your first commit and apologies for not noticing it earlier. Do you know if any of these methods will work to do this: https://stackoverflow.com/questions/3840762/how-do-you-urlencode-without-using-system-web Using System.Web will be a blocker for us moving to asp.net core later this year.

edwardmeng: I have adopted it to Uri.EscapeDataString method to instead of HttpUtility.UrlEncode. And I have learned it. Thanks very much. We are waiting for your moving to asp.net core.

niemyjski:
Thanks for the PR!

PR #318, opened 2017-09-08
Fix bug: The projects cannot be retrieved when there are more than 100 projects in account.

edwardmeng:
Please refer to exceptionless/Exceptionless.UI#154


niemyjski:
I need to talk on this more but I don't think we'll be accepting this as i think this should be fixed client side. Going more than a 100 records per page which is generous can be brutal on the server and networks.

niemyjski:
I'm going to close this, we need to fix this client side with paging.

PR #320, opened 2017-09-14
Add configuration settings parameter for the work timeout of background queues.

edwardmeng:
When the performance of elasticsearch is not fast as expected or there are too much log events, the hard code work timeout(5 minutes) will be not enough to post events to elasticsearch. So expose the QueueWorkTimeout configuration parameter to make chance to manually define the customized value in the deployment stage.


niemyjski:
Could you let me know what is taking so long, the queue work item timeout of 5 minutes is crazy long. It's the period of time from when you dequeue an event post work item to when it completes. Are you seeing a posting of a single event timings out after 5 minutes of it getting processed?

edwardmeng:
Sorry, it's my misunderstanding. I supposed the 5 minutes is started from when enqueue an event post to when it completes.

niemyjski:
I think you may need to scale out your event post jobs. This setting only affects the timeout from when a job dequeues an item to when it's completed. It would not affect enqueue to the time it's dequeued and processed, that would not cause a timeout. Did this change have any other improvements for you? I'd rather not make this change unless there is a real issue here.

Issue #336, opened 2017-12-21
hoping to provide a mechanism to set OccurrencesAreCritical


majian159:
The default does not repeat the notification.

We hope exceptions will always be notified via webhook, but 'OccurrencesAreCritical' defaults to false, which will have a lot of work to set for 'Future Occurrences Are Critical'.

niemyjski:
Hello @majian159, thanks for opening an issue. Can you please explain what your goal is. I'm assuming you are wanting to get notified of every single occurrence that comes in? If that is the case then marking them as critical is not the solution.

majian159:
@niemyjski Thanks, I will try it soon.

niemyjski:
I'm curious, any reason you want a web hook notification for every occurrence?

majian159:
Yes, the cost of accessing webhooks is very low, and we want to notify every error.

niemyjski:
They aren't really low cost unless the server being called has no timeouts, and is low latency and not a very high throughput of events coming in (but still overhead involved). I'm just curious what is being done with every web hook occurrence.

majian159:
Send an error message to an IM software,It doesn't matter if it fails.

ejsmith:
I’m guessing @majian159 doesn’t get a lot of events and that is why he wants notification on all of them. It seems pretty reasonable for small projects or ones that are sending only select events or just don’t have many exceptions.

majian159:
Yeah, one day we missed a lot of timeout exceptions (redis, sql and more) and we think every exception is important.

niemyjski:
Yeah, I get wanting that just was curious how it was being used so we can get a better idea of usages and take that into consideration when working on features/functionality. It would be nice to open this up a bit and give rate based notifications / followup on exceptions once a day.

majian159:
This is very good.

niemyjski:
I'm closing this issue as this can be achieved by wiring up to the Submitting Event handler on the Exceptionless client and mark the event as critical, if you need any help doing this I'd be more than happy to help. I'm also going to update the notifications issue to make a note about web hooks and getting notified more frequently.

PR #338, opened 2017-12-22
Set OccurrencesAreCritical through Project Configuration

majian159:
#336 Set OccurrencesAreCritical through Project Configuration


niemyjski:
Thanks for the PR. I'm closing this issue as this can be achieved by wiring up to the Submitting Event handler on the Exceptionless client and mark the event as critical, if you need any help doing this I'd be more than happy to help. I'm also going to update the notifications issue to make a note about web hooks and getting notified more frequently.

Issue #328, opened 2017-11-06
Improve implementation of high concurrent elasticsearch operations instead of `RetryOnConflict`


edwardmeng:
In high concurrent environment, the option `RetryOnConflict` for elastic operations is not reliable and there will be errors reported as following:
`
StackRepository Error occurred incrementing total event occurrences on stack "59cc6fe3bff2a3c6f00d7961". Error: Type: version_conflict_engine_exception Reason: "[stacks][59cc6fe3bff2a3c6f00d7961]: version conflict, current version [8360562] is different than the one provided [8360561]"
`
The RetryOnConflict has double-edges, although it can improve the success rate, but at the same time it will slow down the elasticsearch updating performance.
more information refer to: https://github.com/exceptionless/Exceptionless/commit/de17ef8ca10d98552e43e16c8d72a990b2419f77

ejsmith:
Yeah, I'd love to see a PR taking a shot at this. I think we could just do something using the cache client to increment a value and then sync it to ES every X seconds. Something similar to what we are doing here:

https://github.com/exceptionless/Exceptionless/blob/master/src/Exceptionless.Core/Services/UsageService.cs

edwardmeng:
Thanks for your reference. I have noticed that in the UsageService, the usage count have been incremented in ES and cache at the same time, but not sync cache usage to ES every X seconds. So it improved the reading performance for the `GetUsage` method, but not the writing performance of the `IncrementUsage` method.

niemyjski:
We increment in cache only (IncrementIf's) and populate the model from that. https://github.com/exceptionless/Exceptionless/blob/master/src/Exceptionless.Core/Services/UsageService.cs#L109 shows we check to see if it should be saved on the repository too. I'm not sure on your last sentence, we only save if usage just exceeded the limit or the last save was 5 minutes ago.

It would be nice to have some kind pattern here (sounds like a saga) that could be reused that basically says, for this giving time period of x, increment the stats and persist it at the end of this period or if this metric / value exceeds this threshold. It would be awesome if this could service restarts and handle saves of old data (but not required for first pass). The usage is a hot code path too and if we can optimize this (we've tried to make it faster and faster every now and then), the difference is the stack count needs to be saved more often as stats in the UI will reflect poorly. I'm fine with the grids being wrong but perhaps the stacks/id end point needs to pull updated metrics / we need to see how those stats are being loaded and pull those from cache.

edwardmeng:
[...]

PR #342, opened 2017-12-28
 Do not publish messages in event post jobs when the SignalR has been disabled

edwardmeng:
We have deployed the exceptionless server with 4 parts: 
1. website cluster: includes multiple exceptionless web servers and nginx for load balance
2. event queue: includes redis and shared file server
3. job cluster: includes multiple servers only for event post jobs
4. elasticsearch cluster: includes multiple elasticsearch shard servers and kibana service.

When the job cluster processing events, there will be at least one message(type ExtendedEntityChanged) published to redis queue, and then every website servers will be received and processed this message. Imagine this scenario:
1. There is no events submitted to the website cluster.
2. There is no users to query events
3. There is lots of events in the event queue.

Then the job cluster servers will be works hard to push the events to elasticsearch, at the same time publish messages to the redis message bus. So the website cluster servers cannot be idle(about 60% cpu usage), no matter whether there is client request or viewers. We got the following log messages again and again for every single event processing
```
2017-12-28 09:51:47.2233   26 TRACE RedisMessageBus OnMessage(messages) 
2017-12-28 09:51:47.2233   30 TRACE RedisMessageBus Found 1 subscribers for message type ExtendedEntityChanged. 
2017-12-28 09:51:47.2233   30 TRACE MessageBusBroker Sending PersistentEvent message to organization: 5a421a98a066d613581f5b2c) 
```

After some investigation, we found that set the `EnableSignalR` to be false can break this nightmare. The the cpu usage decreased to 30%, and we got the following log messages for every single event processing
```
2017-12-28 11:13:47.6867   63 TRACE RedisMessageBus OnMessage(messages) 
2017-12-28 11:13:47.6867   63 TRACE RedisMessageBus Done sending message to 0 subscribers for message type ExtendedEntityChanged. 
```
So we have done the changes that not to publish messages when the SignalR has been disabled to avoid message processing both in job clusters and website clusters.

### DIFF
--- a/src/Exceptionless.Core/Repositories/Base/RepositoryBase.cs
+++ b/src/Exceptionless.Core/Repositories/Base/RepositoryBase.cs
@@ -20,7 +20,8 @@ namespace Exceptionless.Core.Repositories {
         public RepositoryBase(IIndexType<T> indexType, IValidator<T> validator) : base(indexType, validator) {}
 
         protected override Task PublishChangeTypeMessageAsync(ChangeType changeType, T document, IDictionary<string, object> data = null, TimeSpan? delay = null) {
-          return PublishMessageAsync(new ExtendedEntityChanged {
+            if (!Settings.Current.EnableSignalR) return Task.CompletedTask;
+            return PublishMessageAsync(new ExtendedEntityChanged {
                 ChangeType = changeType,
                 Id = document?.Id,
                 OrganizationId = (document as IOwnedByOrganization)?.OrganizationId,
@@ -32,7 +33,7 @@ namespace Exceptionless.Core.Repositories {
         }
 
         protected override async Task SendQueryNotificationsAsync(ChangeType changeType, IRepositoryQuery query, ICommandOptions options) {
-            if (!NotificationsEnabled || !options.ShouldNotify())
+            if (!NotificationsEnabled || !options.ShouldNotify() || !Settings.Current.EnableSignalR)
                 return;
 
             var delay = TimeSpan.FromSeconds(1.5);


niemyjski:
I just merged in asp.net core support as we've been running it in production for a week without any issues (We still have work todo on release artifacts and I'm not sure if we also want to tackle elasticsearch 6.0 since this is going to be a major release). Can you please update your pr / create a new one as this one is conflicted (Please see discussion below).

So we have a new property called `EnableWebSockets` which we are setting to false in our collectors, this just saves some overhead of getting messages on the collector instances and the like. We've also done a TON of work and identified some severe performance issues which have been addressed in master with the redis message bus, which mitigate this completely. For your scenario I can see how there would be some extra load (which is probably completely resolved in the master branch). With that said, if you turn it off on the jobs, the ui will never get updated because it's not getting any messages from the backend jobs and that really stinks. There is no way to know that you don't have subscribers on the backend before publishing change messages but it should be really efficient now as we are using a dedicated task queue that scales out.

edwardmeng:
In my opinion, we should give up some functional supports for the high performance requirements scenario, such as the notifications from jobs to the client users. But for the scenarios without the performance requirements, the functional supports should be applied for user friendly or convinent.
So the `EnableWebSockets` should be `false` for the high performance requirements scenario, and tell the client user to refresh browsers to get the latest data. That is what we intended to do.

niemyjski:
@ejsmith do you have any thoughts on this???? 

I'm wondering if we should create a new property (`EnableChangeNotifications`) and default it to the value of EnableWebSockets. For us we want to keep on the change notifications but still get notified. I hate having extra settings but I could see why you'd want this. I really wonder if asp.net core would mitigate these issues. We did a ton to resolve issues like this but there is a bit of work left to make it more efficient (We need to write custom serializers https://github.com/neuecc/Utf8Json/issues/38)  so we can switch to message pack / utf8Json) for super fast serialization.

edwardmeng:
@niemyjski @ejsmith I don't like extra settings too, when you get agreement, please let me know and I will make changes for the new pr.

ejsmith:
Hmm... I wonder if we could make a massive perf improvement by adding a callback to check to see if a change notification message should be sent or not. This could be wired up to look at a live list of org ids for users that are currently connected to and want real-time notifications. So effectively you would only pay the price of real-time notifications for a small list of org ids that are connected to websockets right now.

edwardmeng:
@ejsmith I think that would be nice to add an option for organization settings page General tab for the Exceptionless.UI, maybe 'Receive notifications' and default to be true. And then we can check this option value that whether send notification messages in the jobs.

ejsmith:
@edwardmeng so you don't want real-time updates in your use case?

What I was talking about would eliminate the work, but only if nobody is currently logged into the app.

edwardmeng:
@ejsmith Comparing the performance and user friendly real-time updates, we will choose the performance and give up user friendly.
Just reviewed the exceptionless website, I think this option should be located in the `My Account -> General` view. Image that we have multiple organizations for my account, but only one of them don't need real-time notification, it will be confused when I selected `All projects` on the header.

ejsmith:
So basically if you have a project that has a very high rate of events then you want to disable real-time notifications for just that project?

I think we really need to spend some time digging in to see if its possible to greatly reduce the number of messages being sent even in cases where a project has a ton of events coming through. I am pretty sure that its possible for us to still have real-time notifications in these scenarios. I know right now we are trying to send the notifications in batches, but I'm thinking maybe we have some messages that are out of control.

niemyjski:
I think if we checked for subscribers and continue trying to get over to message pack format.. Then that would make a huge difference. After that's implemented you could take it a step further with deduping change notifications within a short window per app instance (no cost of network) with observables/rx? You could also have a project or org level setting that short circuited all of this that could be behind an admin route?

edwardmeng:
maybe my fix for the https://github.com/exceptionless/Exceptionless/issues/328 could reduce the amount of notifications.

PR #343, opened 2017-12-28
Migrated to ASP.NET Core

niemyjski:


src/Exceptionless.Core/Extensions/AppDomainExtensions.cs
```diff
@@ -4,7 +4,7 @@
 namespace Exceptionless.Core.Extensions {
     public static class AppDomainExtensions {
         public static void SetDataDirectory(this AppDomain appDomain) {
-            var path = Path.Combine(appDomain.BaseDirectory, @"..\..\..\..\Exceptionless.Api\App_Data");
+            string path = Path.GetFullPath(Path.Combine(appDomain.BaseDirectory, @"..\..\..\..\..\Exceptionless.Web\App_Data"));
```
niemyjski: This probably needs to be using Path.Combine for xplat

PR #349, opened 2018-01-04
Increment stack event count with separate job and run every 5 seconds instead of every event processed.

edwardmeng:
Fix the issue https://github.com/exceptionless/Exceptionless/issues/328, and the pr can reduce the notification messages of the issue https://github.com/exceptionless/Exceptionless/pull/342

### DIFF
--- a/src/Exceptionless.Core/Jobs/StackEventCountJob.cs
+++ b/src/Exceptionless.Core/Jobs/StackEventCountJob.cs
@@ -1,0 +1,33 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Exceptionless.Core.Extensions;
+using Exceptionless.Core.Services;
+using Foundatio.Caching;
+using Foundatio.Jobs;
+using Foundatio.Lock;
+using Microsoft.Extensions.Logging;
+
+namespace Exceptionless.Core.Jobs {
+    [Job(Description = "Update event occurrence count for stacks.", InitialDelay = "2s", Interval = "5s")]
+    public class StackEventCountJob : JobWithLockBase {
+        private readonly StackService _stackService;
+        private readonly ILockProvider _lockProvider;
+
+        public StackEventCountJob(StackService stackService, ICacheClient cacheClient, ILoggerFactory loggerFactory = null) : base(loggerFactory) {
+            _stackService = stackService;
+            _lockProvider = new ThrottlingLockProvider(cacheClient, 1, TimeSpan.FromSeconds(5));
+        }
+
+        protected override async Task<JobResult> RunInternalAsync(JobContext context) {
+            _logger.LogTrace("Start save stack event counts.");
+            await _stackService.SaveStackUsagesAsync(cancellationToken: context.CancellationToken).AnyContext();
+            _logger.LogTrace("Finished save stack event counts.");
+            return JobResult.Success;
+        }
+
+        protected override Task<ILock> GetLockAsync(CancellationToken cancellationToken = default) {
+            return _lockProvider.AcquireAsync(nameof(StackEventCountJob), TimeSpan.FromSeconds(5), new CancellationToken(true));
+        }
+    }
+}
--- a/src/Exceptionless.Core/Services/StackService.cs
+++ b/src/Exceptionless.Core/Services/StackService.cs
@@ -14,6 +14,7 @@ namespace Exceptionless.Core.Services {
         private readonly ILogger<UsageService> _logger;
         private readonly IStackRepository _stackRepository;
         private readonly ICacheClient _cache;
+        private readonly TimeSpan _expireTimeout = TimeSpan.FromHours(12);
 
         public StackService(IStackRepository stackRepository, ICacheClient cache, ILoggerFactory loggerFactory = null) {
             _stackRepository = stackRepository;
@@ -40,13 +41,13 @@ namespace Exceptionless.Core.Services {
 
             var cachedOccurrenceMinDateUtc = await _cache.GetAsync<DateTime>(occurrenceMinDateCacheKey).AnyContext();
             if (!cachedOccurrenceMinDateUtc.HasValue || cachedOccurrenceMinDateUtc.Value.IsAfter(minOccurrenceDateUtc))
-                tasks.Add(_cache.SetAsync(occurrenceMinDateCacheKey, minOccurrenceDateUtc));
+                tasks.Add(_cache.SetAsync(occurrenceMinDateCacheKey, minOccurrenceDateUtc, _expireTimeout));
 
             var cachedOccurrenceMaxDateUtc = await _cache.GetAsync<DateTime>(occurrenceMaxDateCacheKey).AnyContext();
             if (!cachedOccurrenceMaxDateUtc.HasValue || cachedOccurrenceMaxDateUtc.Value.IsBefore(maxOccurrenceDateUtc))
-                tasks.Add(_cache.SetAsync(occurrenceMaxDateCacheKey, maxOccurrenceDateUtc));
+                tasks.Add(_cache.SetAsync(occurrenceMaxDateCacheKey, maxOccurrenceDateUtc, _expireTimeout));
 
-            tasks.Add(_cache.IncrementAsync(occurenceCountCacheKey, count));
+            tasks.Add(_cache.IncrementAsync(occurenceCountCacheKey, count, _expireTimeout));
             tasks.Add(_cache.SetAddAsync(occurrenceSetCacheKey, Tuple.Create(organizationId, projectId, stackId)));
 
             await Task.WhenAll(tasks).AnyContext();
@@ -63,22 +64,17 @@ namespace Exceptionless.Core.Services {
                 string occurrenceCountCacheKey = GetStackOccurrenceCountCacheKey(organizationId, projectId, stackId),
                     occurrenceMinDateCacheKey = GetStackOccurrenceMinDateCacheKey(organizationId, projectId, stackId),
                     occurrenceMaxDateCacheKey = GetStackOccurrenceMaxDateCacheKey(organizationId, projectId, stackId);
-                var occurrenceCount = _cache.GetAsync<long>(occurrenceCountCacheKey, 0);
+                var occurrenceCountTask = _cache.GetAsync<long>(occurrenceCountCacheKey, 0);
                 var occurrenceMinDate = _cache.GetAsync(occurrenceMinDateCacheKey, SystemClock.UtcNow);
                 var occurrenceMaxDate = _cache.GetAsync(occurrenceMaxDateCacheKey, SystemClock.UtcNow);
 
-                await Task.WhenAll(occurrenceCount, occurrenceMinDate, occurrenceMaxDate).AnyContext();
-                if (occurrenceCount.Result == 0) continue;
-                await _cache.RemoveAllAsync(new[] { occurrenceCountCacheKey, occurrenceMinDateCacheKey, occurrenceMaxDateCacheKey }).AnyContext();
+                await Task.WhenAll(occurrenceCountTask, occurrenceMinDate, occurrenceMaxDate).AnyContext();
+                var occurrenceCount = (int)occurrenceCountTask.Result;
+                if (occurrenceCount == 0) continue;
 
-                try {
-                    await _stackRepository.IncrementEventCounterAsync(organizationId, projectId, stackId, occurrenceMinDate.Result, occurrenceMaxDate.Result, (int)occurrenceCount.Result, sendNotifications).AnyContext();
-                    _logger.LogTrace("Increment event count {occurrenceCount} for organization:{organizationId} project:{projectId} stack:{stackId} with occurrenceMinDate:{occurrenceMinDate} occurrenceMaxDate:{occurrenceMaxDate}", occurrenceCount.Result, organizationId, projectId, stackId, occurrenceMinDate.Result, occurrenceMaxDate.Result);
-                }
-                catch {
-                    await IncrementStackUsageAsync(organizationId, projectId, stackId, occurrenceMinDate.Result, occurrenceMaxDate.Result, (int)occurrenceCount.Result).AnyContext();
-                    throw;
-                }
+                await _stackRepository.IncrementEventCounterAsync(organizationId, projectId, stackId, occurrenceMinDate.Result, occurrenceMaxDate.Result, occurrenceCount, sendNotifications).AnyContext();
+                await _cache.DecrementAsync(occurrenceCountCacheKey, occurrenceCount, _expireTimeout).AnyContext();
+                _logger.LogTrace("Increment event count {occurrenceCount} for organization:{organizationId} project:{projectId} stack:{stackId} with occurrenceMinDate:{occurrenceMinDate} occurrenceMaxDate:{occurrenceMaxDate}", occurrenceCount, organizationId, projectId, stackId, occurrenceMinDate.Result, occurrenceMaxDate.Result);
             }
         }
 
--- a/src/Exceptionless.Core/Services/StackService.cs
+++ b/src/Exceptionless.Core/Services/StackService.cs
@@ -1,10 +1,8 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Exceptionless.Core.Extensions;
 using Exceptionless.Core.Repositories;
-using Exceptionless.DateTimeExtensions;
 using Foundatio.Caching;
 using Foundatio.Utility;
 using Microsoft.Extensions.Logging;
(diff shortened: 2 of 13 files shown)


edwardmeng:
I have completed this pr, please take review.

src/Exceptionless.Core/Jobs/StackEventCountJob.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        protected override async Task<JobResult> RunInternalAsync(JobContext context) {
+            _logger.LogTrace("Start save stack event counts.");
+            await _stackService.SaveStackUsagesAsync(cancellationToken: context.CancellationToken).AnyContext();
+            _logger.LogTrace("Finished save stack event counts.");
+            return JobResult.Success;
+        }
+
+        protected override Task<ILock> GetLockAsync(CancellationToken cancellationToken = default) {
+            return _lockProvider.AcquireAsync(nameof(StackEventCountJob), TimeSpan.FromSeconds(5), new CancellationToken(true));
```
niemyjski: One concern I have about this is that if you have a large number of stacks to run (we do), you can't distribute the job. But you also don't know what stack it's for because it's a generic job :\. We should think about this some more.

src/Exceptionless.Core/Services/StackService.cs
```diff
@@ -0,0 +1,94 @@
...
+        private readonly ICacheClient _cache;
+
+        public StackService(IStackRepository stackRepository, ICacheClient cache, ILoggerFactory loggerFactory = null) {
+            _stackRepository = stackRepository;
+            _cache = cache;
+            _logger = loggerFactory.CreateLogger<UsageService>();
+        }
+
+        public async Task IncrementStackUsageAsync(string organizationId, string projectId, string stackId, DateTime minOccurrenceDateUtc, DateTime maxOccurrenceDateUtc, int count) {
+            if (String.IsNullOrEmpty(organizationId) || String.IsNullOrEmpty(projectId) || String.IsNullOrEmpty(stackId) || count == 0)
```
niemyjski: Should probably throw a ArgumentNullException if any of these strings are null and bail if the count is less than or equal to 0

src/Exceptionless.Core/Services/StackService.cs
```diff
@@ -0,0 +1,94 @@
...
+                occurrenceMinDateCacheKey = GetStackOccurrenceMinDateCacheKey(organizationId, projectId, stackId),
+                occurrenceMaxDateCacheKey = GetStackOccurrenceMaxDateCacheKey(organizationId, projectId, stackId),
+                occurrenceSetCacheKey = GetStackOccurrenceSetCacheKey();
+
+            var cachedOccurrenceMinDateUtc = await _cache.GetAsync<DateTime>(occurrenceMinDateCacheKey).AnyContext();
+            if (!cachedOccurrenceMinDateUtc.HasValue || cachedOccurrenceMinDateUtc.IsNull || cachedOccurrenceMinDateUtc.Value > minOccurrenceDateUtc)
+                tasks.Add(_cache.SetAsync(occurrenceMinDateCacheKey, minOccurrenceDateUtc, expireTimeout));
+
+            var cachedOccurrenceMaxDateUtc = await _cache.GetAsync<DateTime>(occurrenceMaxDateCacheKey).AnyContext();
+            if (!cachedOccurrenceMaxDateUtc.HasValue || cachedOccurrenceMinDateUtc.IsNull || cachedOccurrenceMaxDateUtc.Value < maxOccurrenceDateUtc)
```
niemyjski: Not sure if you have to do HasValue and IsNull, I think if it has a value it should never be null. We have date time extensions for .IsBefore and .IsAfter that makes this a bit easier to read (but it's fine)

src/Exceptionless.Core/Services/StackService.cs
```diff
@@ -0,0 +1,94 @@
...
+                tasks.Add(_cache.SetAsync(occurrenceMaxDateCacheKey, maxOccurrenceDateUtc, expireTimeout));
+
+            tasks.Add(_cache.IncrementAsync(occurenceCountCacheKey, count, expireTimeout));
```
niemyjski: I'm really not sure if we should be expiring every one of these cache keys, I mean it's a good thing but under the covers it's doing two operations (one for set, and one for expire I believe).

src/Exceptionless.Core/Services/StackService.cs
```diff
@@ -0,0 +1,94 @@
...
+            tasks.Add(_cache.SetAddAsync(occurrenceSetCacheKey, Tuple.Create(organizationId, projectId, stackId), expireTimeout));
```
niemyjski: I'm not sure but I think setting expire on the list will expire the whole list and not just the individual item. I'm wondering if we even have tests for this specific scenario in foundatio.

src/Exceptionless.Core/Services/StackService.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            await Task.WhenAll(tasks).AnyContext();
+        }
+
+        public async Task SaveStackUsagesAsync(bool sendNotifications = true, CancellationToken cancellationToken = default) {
+            var occurrenceSetCacheKey = GetStackOccurrenceSetCacheKey();
+            var stackUsageSet = await _cache.GetSetAsync<Tuple<string, string, string>>(occurrenceSetCacheKey).AnyContext();
```
niemyjski: I'm curious, do you know how Tuples round trips via binary serialization?

src/Exceptionless.Core/Services/StackService.cs
```diff
@@ -0,0 +1,94 @@
...
+            if (!stackUsageSet.HasValue || stackUsageSet.IsNull) return;
+            foreach (var tuple in stackUsageSet.Value) {
+                if (cancellationToken.IsCancellationRequested) break;
+
+                string organizationId = tuple.Item1, projectId = tuple.Item2, stackId = tuple.Item3;
+                string occurenceCountCacheKey = GetStackOccurrenceCountCacheKey(organizationId, projectId, stackId),
+                    occurrenceMinDateCacheKey = GetStackOccurrenceMinDateCacheKey(organizationId, projectId, stackId),
+                    occurrenceMaxDateCacheKey = GetStackOccurrenceMaxDateCacheKey(organizationId, projectId, stackId);
+                var occurrenceCount = await _cache.GetAsync<long>(occurenceCountCacheKey, 0).AnyContext();
```
niemyjski: I wish we could do a get and decrement at the same time (atomically), I feel like this could be a race condition long term (but will be eventually consistent)

src/Exceptionless.Core/Services/StackService.cs
```diff
@@ -0,0 +1,94 @@
...
+                if (occurrenceCount == 0) return;
+                var occurrenceMinDate = _cache.GetAsync(occurrenceMinDateCacheKey, SystemClock.UtcNow);
+                var occurrenceMaxDate = _cache.GetAsync(occurrenceMaxDateCacheKey, SystemClock.UtcNow);
+
+                await Task.WhenAll(occurrenceMinDate, occurrenceMaxDate).AnyContext();
+
+                var tasks = new List<Task> {
+                    _stackRepository.IncrementEventCounterAsync(organizationId, projectId, stackId, occurrenceMinDate.Result, occurrenceMaxDate.Result, (int)occurrenceCount, sendNotifications),
+                    _cache.RemoveAllAsync(new[] { occurenceCountCacheKey, occurrenceMinDateCacheKey, occurrenceMaxDateCacheKey }),
+                    _cache.SetRemoveAsync(occurrenceSetCacheKey, tuple, TimeSpan.FromHours(24))
```
niemyjski: We gotta think about this as I know we can't do it this way due to race conditions. We could be removing values even from the time we were getting values from the cache (that were being added after we got our value). We gotta think of a way to do this without locking on incrementing as that would slow down the pipeline lots too. 

Not sure what the FromHours(24) is here. I wonder if we could just pop an item from a list and get the value and then remove that, that way we know anything in that list was old? @ejsmith @edwardmeng thoughts?

src/Exceptionless.Core/Services/StackService.cs
```diff
@@ -0,0 +1,94 @@
...
+                };
+                await Task.WhenAll(tasks).AnyContext();
+
+                _logger.LogTrace($"Increment event count {occurrenceCount} for organization:{organizationId} project:{projectId} stack:{stackId} with occurrenceMinDate:{occurrenceMinDate.Result} occurrenceMaxDate:{occurrenceMaxDate.Result}");
```
niemyjski: This should use our structured logging so we don't take the penalty of calculating the string if the logger isn't enabled. (just gotta do Friendly names in the brackets and then pass the argument to the end `LogTrace(.."count {OccurrenceCount}.......", occurrenceCount);`

niemyjski:
I'm also thinking we could use this same pattern once profected to save usage stats so the pipeline isn't doing as much.

ejsmith:
I agree with @niemyjski. I like the general direction here, but I'm concerned about race conditions between the job collecting and resetting the stats. So you have a job that is grabbing the set of stack ids and then updating the docs and then removing the item from the set. But at the same time you could have that exact same stack id being added back to the set. I think the way to avoid this would be to create a LUA script that would take a stack id, get the current count from it and remove it from the set. Since Redis will run that script as a single operation then it would avoid the race condition.

niemyjski:
But then you are very dependent on redis for this to work. Perhaps we had
new ability to caches?

Thanks
-Blake Niemyjski

ejsmith:
Yep, that is true. Not sure how else to solve it. Guess the only way to solve it if we don't want to require Redis is to just use our locks which sucks because the lock will need to be used from the pipeline as well which means we are adding a huge source of contention to a critical piece of the code.

src/Exceptionless.Core/Services/StackService.cs
edwardmeng: Actually I don't know whether it can be null, but check it for safety concern to avoid unexpected scenarios.

src/Exceptionless.Core/Services/StackService.cs
edwardmeng: I'm not sure about the expiring operation too, but most similar cache operations are using it, so I followed and provided the expire timeout 1d that long enough to processing for the task.

src/Exceptionless.Core/Services/StackService.cs
edwardmeng: The Tuples will be serialized same as POCO

src/Exceptionless.Core/Services/StackService.cs
edwardmeng: When saving stack usages, we will retrieve the count of the events in the period between last saving and this saving and increment this number to elasticsearch, then we should remove the occurrence count or set it 0 for the incoming event post job processing.  
Suppose we want to achieve this purpose through the `decrement` operation, we have to know the occurrence count before operation.
I think the better choice is: Add new method for the `ICacheClient` that can get and set cache value at the same time atomically. As I know there is API for this operation
https://github.com/StackExchange/StackExchange.Redis/blob/c4c9c1fdb455070415e82d2f104fc89a90b057b5/StackExchange.Redis/StackExchange/Redis/IDatabase.cs#L1020
and
https://github.com/StackExchange/StackExchange.Redis/blob/c4c9c1fdb455070415e82d2f104fc89a90b057b5/StackExchange.Redis/StackExchange/Redis/IDatabaseAsync.cs#L960

src/Exceptionless.Core/Jobs/StackEventCountJob.cs
edwardmeng: I think when the race conditions problem fixed, the lock should be removed.

edwardmeng:
I have completed changes, please review again.

edwardmeng:
I have completed the changes, please review again.

niemyjski:
Looks like there is a merge conflict on the solution file :\

src/Exceptionless.Core/Services/StackService.cs
ejsmith: It looks like we are only adding to the set and never clearing it out which means it will have unbounded growth and this call will end up being a massive amount of data.

ejsmith: I am thinking that maybe we can get the full set. Get the current count of the stack. Remove the item from the set. And then decrement occurrence count by the value that we had. Then even if at the same time the stack is being added back to the set it will be fine because it will be incrementing the count and we will be decrementing it. This approach requires us to add a SetRemove operation to the cache client.

src/Exceptionless.Core/Services/StackService.cs
```diff
@@ -0,0 +1,97 @@
...
+                return;
+
+            var tasks = new List<Task>(4);
+
+            string occurenceCountCacheKey = GetStackOccurrenceCountCacheKey(organizationId, projectId, stackId),
+                occurrenceMinDateCacheKey = GetStackOccurrenceMinDateCacheKey(organizationId, projectId, stackId),
+                occurrenceMaxDateCacheKey = GetStackOccurrenceMaxDateCacheKey(organizationId, projectId, stackId),
+                occurrenceSetCacheKey = GetStackOccurrenceSetCacheKey();
+
+            var cachedOccurrenceMinDateUtc = await _cache.GetAsync<DateTime>(occurrenceMinDateCacheKey).AnyContext();
```
ejsmith: This should be using the SetIfHigherAsync and using the long epoch value so that we don't have to do a get and a set which could also be a race condition.

src/Exceptionless.Core/Services/StackService.cs
```diff
@@ -0,0 +1,97 @@
...
+            if (!cachedOccurrenceMinDateUtc.HasValue || cachedOccurrenceMinDateUtc.Value.IsAfter(minOccurrenceDateUtc))
+                tasks.Add(_cache.SetAsync(occurrenceMinDateCacheKey, minOccurrenceDateUtc, _expireTimeout));
+
+            var cachedOccurrenceMaxDateUtc = await _cache.GetAsync<DateTime>(occurrenceMaxDateCacheKey).AnyContext();
+            if (!cachedOccurrenceMaxDateUtc.HasValue || cachedOccurrenceMaxDateUtc.Value.IsBefore(maxOccurrenceDateUtc))
+                tasks.Add(_cache.SetAsync(occurrenceMaxDateCacheKey, maxOccurrenceDateUtc, _expireTimeout));
```
ejsmith: This should be using the SetIfLowerAsync and using the long epoch value so that we don't have to do a get and a set which could also be a race condition.

src/Exceptionless.Core/Services/StackService.cs
ejsmith: Not sure if we have a utility method for this some place, but here is some code if we need it:
https://stackoverflow.com/a/3354915/55890

edwardmeng:
@niemyjski @ejsmith I have completed the changes, please take review again.

src/Exceptionless.Core/Services/StackService.cs
```diff
@@ -0,0 +1,113 @@
...
+                }
+
+                Task IncrementStackUsage() {
+                    return IncrementStackUsageAsync(organizationId, projectId, stackId, occurrenceMinDate, occurrenceMaxDate, occurrenceCount);
+                }
+            }
+        }
+
+        internal double ConvertToUnixTimeStamp(DateTime dateUtc) {
+            return (dateUtc - _unixTimeStampOrigin).Ticks;
```
niemyjski: Shouldn't we be able to just use the dateUtc ticks? instead of converting it every time? Or was there some kind of overflow?

src/Exceptionless.Core/Services/StackService.cs
```diff
@@ -0,0 +1,113 @@
...
+                var occurrenceMinDateTask = _cache.GetAsync<double>(occurrenceMinDateCacheKey);
+                var occurrenceMaxDateTask = _cache.GetAsync<double>(occurrenceMaxDateCacheKey);
+
+                await Task.WhenAll(occurrenceCountTask, occurrenceMinDateTask, occurrenceMaxDateTask).AnyContext();
+                var occurrenceCount = (int)occurrenceCountTask.Result;
+                if (occurrenceCount == 0) continue;
+
+                await Task.WhenAll(
+                    _cache.DecrementAsync(occurrenceCountCacheKey, occurrenceCount, _expireTimeout),
+                    _cache.SetRemoveAsync(occurrenceSetCacheKey, tuple),
```
niemyjski: I think this set remove task should be moved up to the previous when all on line 57 to further prevent a race condition.

src/Exceptionless.Core/Services/StackService.cs
```diff
@@ -0,0 +1,113 @@
...
+                try {
+                    if (!await _stackRepository.IncrementEventCounterAsync(organizationId, projectId, stackId, occurrenceMinDate, occurrenceMaxDate, occurrenceCount, sendNotifications).AnyContext()) {
+                        await IncrementStackUsage().AnyContext();
+                    }
+                    else {
+                        _logger.LogTrace("Increment event count {occurrenceCount} for organization:{organizationId} project:{projectId} stack:{stackId} with occurrenceMinDate:{occurrenceMinDate} occurrenceMaxDate:{occurrenceMaxDate}", occurrenceCount, organizationId, projectId, stackId, occurrenceMinDate, occurrenceMaxDate);
+                    }
+                }
+                catch {
+                    await IncrementStackUsage().AnyContext();
```
niemyjski: We should add logging about what exception was thrown here, I'm also not sure if we should be incrementing the stack usage here. What if the previous IncrementStackUsage() is the one that threw, we could get a double count?

niemyjski:
Looks good, I can make some of these changes if you'd like. @ejsmith do you have anything to add?

src/Exceptionless.Core/Services/StackService.cs
edwardmeng: I tried this at the beginning, but the redis performs wired. when the number is to big, some bits will be lost. So I have to keep the number as small as possible.

edwardmeng: I tried to let `_unixTimeStampOrigin = new DateTime(1970,1,1)`, and the number retrieve from redis will be not accurate too. After let `_unixTimeStampOrigin = new DateTime(2000,1,1)`, all tests passed.

src/Exceptionless.Core/Services/StackService.cs
edwardmeng: Could you please help me to fix this issue?

edwardmeng:
I have done for these changes, please review again.

niemyjski:
Looks good to me, I'm going to merge it into a perf branch and do testings and accumulate a few other changes like api throttle.

PR #351, opened 2018-01-13
Don't throttle when the ApiThrottleLimit is -1 or int.MaxValue

edwardmeng:


src/Exceptionless.Core/Extensions/OrganizationExtensions.cs
```diff
@@ -77,6 +77,7 @@ public static class OrganizationExtensions {
         }
 
         public static async Task<bool> IsOverRequestLimitAsync(string organizationId, ICacheClient cacheClient, int apiThrottleLimit) {
+            if (apiThrottleLimit < 0 || apiThrottleLimit == int.MaxValue) return false;
```
niemyjski: We shouldn't be checking for -1 just Int32.MaxValue; We need to normalize -1 in the settings.

src/Exceptionless.Api/Startup.cs
```diff
@@ -39,8 +39,12 @@ public class Startup {
             app.UseAuthentication();
             app.UseMiddleware<ProjectConfigMiddleware>();
             app.UseMiddleware<RecordSessionHeartbeatMiddleware>();
-            // Throttle api calls to X every 15 minutes by IP address.
-            app.UseMiddleware<ThrottlingMiddleware>();
+
+            if (Settings.Current.ApiThrottleLimit >= 0 && Settings.Current.ApiThrottleLimit != int.MaxValue) {
+                // Throttle api calls to X every 15 minutes by IP address.
```
niemyjski: We shouldn't be checking for -1 just Int32.MaxValue; We need to normalize -1 in the settings.

PR #354, opened 2018-01-29
Add proper swagger support for a versioned api

caesay:
Cleans up the swagger doc around versioned api's. No longer have optional {version} params in swagger spec. No longer have v1 and v2 api's in the same swagger spec. (there are now two exported swagger specs)


niemyjski:
Thanks for the PR

PR #358, opened 2018-02-16
Improve the handling of IPs on request info

niemyjski:
- Previously local ips could be getting stripped from the results
- This only adds the submitting ip address if there are no remote ip addresses in the list or the list is empty.
- Still adds the requesting ip to the global ip list for sorting.
- This doesn't update the geo plugin to look at the submitting ip.... (We should talk about this)

### DIFF
--- a/src/Exceptionless.Core/Models/Data/RequestInfo.cs
+++ b/src/Exceptionless.Core/Models/Data/RequestInfo.cs
@@ -114,6 +114,11 @@ namespace Exceptionless.Core.Models.Data {
             public const string OSMajorVersion = "@os_major_version";
 
             public const string IsBot = "@is_bot";
+
+            /// <summary>
+            /// The IP address that submitted the event to our api.
+            /// </summary>
+            public const string RequestIpAddress = "@request_ip";
         }
     }
 }
--- a/src/Exceptionless.Core/Models/Data/RequestInfo.cs
+++ b/src/Exceptionless.Core/Models/Data/RequestInfo.cs
@@ -114,11 +114,6 @@ namespace Exceptionless.Core.Models.Data {
             public const string OSMajorVersion = "@os_major_version";
 
             public const string IsBot = "@is_bot";
-
-            /// <summary>
-            /// The IP address that submitted the event to our api.
-            /// </summary>
-            public const string RequestIpAddress = "@request_ip";
         }
     }
 }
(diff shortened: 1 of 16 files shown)


src/Exceptionless.Core/Models/Data/RequestInfo.cs
```diff
@@ -114,6 +114,11 @@ public static class KnownDataKeys {
             public const string OSMajorVersion = "@os_major_version";
 
             public const string IsBot = "@is_bot";
+
+            /// <summary>
+            /// The IP address that submitted the event to our api.
+            /// </summary>
+            public const string RequestIpAddress = "@request_ip";
```
ejsmith: I don't understand this. We already have ClientIpAddress. If this is just the IP address of the Exceptionless client that submitted the events to our API then: a) it should not be on RequestInfo and b) we already have that from the HTTP headers of the incoming event post and that gets put into the event payload data.

niemyjski: yeah, this is passed from the client payload data. We put it in the client IP address because most of the time (very large percentage) there is no ip because there is no way to get the remote ip without doing network calls.

ejsmith: So I have no idea why we would need to add it to the RequestInfo. RequestInfo is specifically for the HTTP request that is attached to the event. Not the Exceptionless client request that is submitting the events to our API.

niemyjski: I think we were always trying to tell you what server / machine that the request happened on in case there wasn't an external address.

src/Exceptionless.Core/Plugins/EventProcessor/Default/07_SubmissionClientPlugin.cs
```diff
@@ -0,0 +1,42 @@
...
+using Exceptionless.Core.Models.Data;
+using Exceptionless.Core.Pipeline;
+using Microsoft.Extensions.Logging;
+
+namespace Exceptionless.Core.Plugins.EventProcessor.Default {
+    [Priority(7)]
+    public sealed class SubmissionClientPlugin : EventProcessorPluginBase {
+        public SubmissionClientPlugin(ILoggerFactory loggerFactory = null) : base(loggerFactory) { }
+
+        public override Task EventProcessingAsync(EventContext context) {
```
niemyjski: Guess it would be better to do this in batch mode.

PR #359, opened 2018-02-19
 Integrate with Minio for file storage.

edwardmeng:


caesay:
@niemyjski The way you currently deal with connection strings seems really dumb to me - if the connection string exists it simply replaces the current instance of storage with it's own. It's unclear what would happen if you specified two connection strings at the same time unless you review the code and see. Generally there would be a "StorageProvider" config option which would then be one of "Azure, Aws, Minio, etc". If there's an additional param for a specific provider that isn't captured by the connection string, it would be named separately. So you might have the following config options:

- StorageProviderName
- StorageConnectionString
- AwsSpecialRequiredFieldThatIsntInTheConnectionString

niemyjski:
@caesay We tried to make configuration as easy as possible and one aspect of that is to have less properties and have a more defined approach to implementations in the sense that it will go from local to remote implementation when defined. For Example redis is used in multiple spots so we go in memory, redis and then any specific implementation which would override that.

niemyjski:
Thanks for the PR.

ejsmith:
@caesay yes, we are trying to avoid having 100 config options. I think we can get the best of both worlds though by implementation a change that @edwardmeng and I have been talking about which is to just have a StorageConnectionString setting and part of the connection string would be the implementation type to use. So like "provider=minio;bucket=my-bucket" or "provider=azure;container=my-container". The connection string would have everything needed inside of it to completely configure and use the specified provider and you would only have 1 setting for all providers which would be an even further simplification of the settings that we have now.

PR #363, opened 2018-03-16
Disperse the event files into different sub-folders according to the prefix of file name.

edwardmeng:
When the files count more than 100,000 then performance of hard disk will becomes worse, and there are no more than 10,000 files can get the best result. So disperse the files into different sub-folders can improve the performance and stability.

### DIFF
--- a/src/Exceptionless.Web/Properties/launchSettings.json
+++ b/src/Exceptionless.Web/Properties/launchSettings.json
@@ -12,7 +12,7 @@
       "commandName": "IISExpress",
       "launchBrowser": true,
       "environmentVariables": {
-        "AppMode": "Production"
+        "AppMode": "Development"
       }
     },
     "Exceptionless.Web": {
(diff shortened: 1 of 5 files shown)


niemyjski:
Looks good, bit worried about a ton of folders being created in the low throughput case (36*36*36, wonder if two characters would still provide the benefit). Could you also update GetArchivePath to include the minute in the archive path.

edwardmeng:
Please take review again

src/Exceptionless.Web/Properties/launchSettings.json
```diff
@@ -12,7 +12,7 @@
       "commandName": "IISExpress",
       "launchBrowser": true,
       "environmentVariables": {
-        "AppMode": "Development"
+        "AppMode": "Production"
```
niemyjski: Can you revert this change please :)

edwardmeng:
Please take review again.

niemyjski:
Thanks for the pr

PR #383, opened 2018-08-17
Integrate with AppMetrics and report to InfluxDB.

edwardmeng:
The AppMetrics can provide much more metrics measures, healthchecking and reportings.


edwardmeng:
I have completed this feature, please take a review.

ejsmith:
Can you please rebase this?

src/Exceptionless.Core/Settings.cs
```diff
@@ -69,7 +69,11 @@ public class Settings {
 
         public int MetricsServerPort { get; private set; }
 
-        public bool EnableMetricsReporting { get; private set; }
+        public string MetricsReportingDatabase { get; private set; }
+
+        public string MetricsReportingStrategy { get; private set; }
```
ejsmith: I would prefer that we not have a metrics reporting mode, but we just check for the existence of config settings in order of highest priority. Because I want the least amount of config settings as possible. So I want things to be inferred and maybe even use a connection string format for this grouping of metrics settings so that we could have a single setting that contains all of the metrics config

niemyjski: Would be great to have one for queues and storage as well. That way we don't have to have a ton of settings for extra implementations.

src/Exceptionless.Core/Settings.cs
```diff
@@ -65,11 +65,7 @@ public class Settings {
 
         public string ApplicationInsightsKey { get; private set; }
 
-        public string MetricsServerName { get; private set; }
-
-        public int MetricsServerPort { get; private set; }
-
-        public bool EnableMetricsReporting { get; private set; }
+        public IMetricsConnectionString MetricsConnectionString { get; private set; }
```
niemyjski: I think this should just be a string type and we need to add the enable metrics reporting back @ejsmith unless you think this should be part of the connection string. This allowed us to add a setting to disable it without touching the connection string and also allowed easy way to see if we enabled it.

src/Exceptionless.Core/Settings.cs
```diff
@@ -250,6 +244,7 @@ public class Settings {
             settings.AzureStorageQueueConnectionString = configRoot.GetConnectionString("AzureStorageQueue");
             settings.AliyunStorageConnectionString = configRoot.GetConnectionString("AliyunStorage");
             settings.MinioStorageConnectionString = configRoot.GetConnectionString("MinioStorage");
+            settings.MetricsConnectionString = Utility.MetricsConnectionString.Parse(configRoot.GetConnectionString("Metrics"));
```
niemyjski: This should just set the connection string and the parsing should happen in the insulation project. The core project shouldn't know anything about statsd or any other implementation.

src/Exceptionless.Core/Utility/MetricsConnectionString.cs
```diff
@@ -0,0 +1,170 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Exceptionless.Core.Utility {
```
niemyjski: I think this belongs in the insulation project, the core project shouldn't know about any concrete implementations, only the abstraction.

edwardmeng:
@niemyjski I have tested the metrics against InfluxDB and Prometheus. Could you please review this pr and help me to test StatsD and Graphite, which I cannot install the environment.

src/Exceptionless.Web/Bootstrapper.cs
```diff
@@ -1,25 +1,55 @@
...
+using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Stripe;
 
 namespace Exceptionless.Web {
     public class Bootstrapper {
+        public static void ConfigureWebHost(IWebHostBuilder builder) {
+            if (!String.IsNullOrEmpty(Settings.Current.ApplicationInsightsKey))
```
niemyjski: This is happening twice and should be removed. Can we just move this to an extension method?

src/Exceptionless.Web/Program.cs
```diff
@@ -73,10 +75,14 @@ public class Program {
...
                 })
                 .UseStartup<Startup>();
-
-            if (useApplicationInsights)
+				
+			if (useApplicationInsights)
                 builder.UseApplicationInsights(Settings.Current.ApplicationInsightsKey);
 
+            if (settings.EnableMetricsReporting && settings.MetricsConnectionString is DefaultMetricsConnectionString) {
+                settings.MetricsConnectionString = MetricsConnectionString.Parse(settings.MetricsConnectionString.ConnectionString);
```
niemyjski: Not sure if we need this check to see if it's the default as it it's the main entry point and doesn't hurt anything to parse It twice if it's by accident.

src/Exceptionless.Insulation/Bootstrapper.cs
```diff
@@ -79,7 +86,8 @@ public class Bootstrapper {
                 container.ReplaceSingleton(s => CreateAzureStorageQueue<WebHookNotification>(s));
                 container.ReplaceSingleton(s => CreateAzureStorageQueue<MailMessage>(s));
                 container.ReplaceSingleton(s => CreateAzureStorageQueue<WorkItemData>(s, workItemTimeout: TimeSpan.FromHours(1)));
-            } else if (!String.IsNullOrEmpty(Settings.Current.RedisConnectionString)) {
+            }
```
niemyjski: Extra formatting changes

src/Exceptionless.Insulation/Metrics/MetricsConnectionString.cs
```diff
@@ -0,0 +1,52 @@
...
+                if (String.Equals(provider, "http", StringComparison.OrdinalIgnoreCase)) {
+                    return new HttpMetricsConnectionString(connectionString, options);
+                }
+
+                if (String.Equals(provider, "influxdb", StringComparison.OrdinalIgnoreCase)) {
+                    return new InfuxDBMetricsConnectionString(connectionString, options);
+                }
+
+                if (String.Equals(provider, "prometheus", StringComparison.OrdinalIgnoreCase)) {
+                    return new PrometheusMetricsConnectionString(connectionString);
```
niemyjski: Why not pass the options here as they've  already been parsed?

PR #393, opened 2018-10-22
Feature/unified config

niemyjski:


src/Exceptionless.Core/Jobs/Elastic/CleanupSnapshotJob.cs
```diff
@@ -5,19 +5,23 @@
...
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace Exceptionless.Core.Jobs.Elastic {
     [Job(Description = "Removes old Elasticsearch snapshots.", IsContinuous = false)]
     public class CleanupSnapshotJob : Foundatio.Repositories.Elasticsearch.Jobs.CleanupSnapshotJob {
-        public CleanupSnapshotJob(ExceptionlessElasticConfiguration configuration, ILockProvider lockProvider, ILoggerFactory loggerFactory)
+        private readonly IOptionsSnapshot<AppOptions> _options;
+
+        public CleanupSnapshotJob(ExceptionlessElasticConfiguration configuration, ILockProvider lockProvider, IOptionsSnapshot<AppOptions> options, ILoggerFactory loggerFactory)
```
ejsmith: Why are you using IOptionsSnapshot? We don't support this changing on the fly while the app is running.

src/Exceptionless.Core/Jobs/EventPostsJob.cs
```diff
@@ -49,6 +51,10 @@ public class EventPostsJob : QueueJobBase<EventPost> {
             _organizationRepository = organizationRepository;
             _projectRepository = projectRepository;
             _jsonSerializerSettings = jsonSerializerSettings;
+            
+            _appOptions = appOptions;
+            _maximumEventPostFileSize = _appOptions.Value.MaximumEventPostSize + 1000;
```
ejsmith: Why are we adding 1000 to this?

src/Exceptionless.Core/Repositories/Configuration/ExceptionlessElasticConfiguration.cs
```diff
@@ -43,7 +47,7 @@ public sealed class ExceptionlessElasticConfiguration : ElasticConfiguration, IS
             builder.Register(new StackQueryBuilder());
         }
 
-        public Settings Settings { get; }
+        public IOptionsSnapshot<ElasticsearchOptions> ElasticsearchOptions { get; }
```
ejsmith: Let's rename this to just Options. It's already inside of a class that is Elasticsearch specific.

src/Exceptionless.Core/Repositories/Configuration/Indexes/OrganizationIndex/OrganizationIndex.cs
```diff
@@ -19,8 +21,8 @@ public sealed class OrganizationIndex : VersionedIndex {
         public override CreateIndexDescriptor ConfigureIndex(CreateIndexDescriptor idx) {
             return base.ConfigureIndex(idx.Settings(s => s
                 .Analysis(d => d.Analyzers(b => b.Custom(KEYWORD_LOWERCASE_ANALYZER, c => c.Filters("lowercase").Tokenizer("keyword"))))
-                .NumberOfShards(_settings.ElasticsearchNumberOfShards)
-                .NumberOfReplicas(_settings.ElasticsearchNumberOfReplicas)
+                .NumberOfShards(_elasticsearchOptions.Value.NumberOfShards)
```
ejsmith: Maybe be less verbose all over the place and just call it _options and know that it is options for the class that we are in. Don't really like us having to access both AppOptions and ElasticsearchOptions. We should just add ScopePrefix to ElasticsearchOptions.

src/Exceptionless.Core/Repositories/EventRepository.cs
```diff
@@ -127,7 +127,7 @@ public EventRepository(ExceptionlessElasticConfiguration configuration, IValidat
             if (ev == null)
                 return null;
 
-            var retentionDate = Settings.Current.MaximumRetentionDays > 0 ? SystemClock.UtcNow.Date.SubtractDays(Settings.Current.MaximumRetentionDays) : DateTime.MinValue;
+            var retentionDate = AppOptions.Current.MaximumRetentionDays > 0 ? SystemClock.UtcNow.Date.SubtractDays(AppOptions.Current.MaximumRetentionDays) : DateTime.MinValue;
```
ejsmith: If these options were added to the Elasticsearch options then you could access them from the configuration that is passed in and not have to use statics for this.

src/Exceptionless.Insulation/Bootstrapper.cs
```diff
@@ -26,162 +26,188 @@
...
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 using Serilog.Sinks.Exceptionless;
 using StackExchange.Redis;
 
 namespace Exceptionless.Insulation {
     public class Bootstrapper {
-        public static void RegisterServices(IServiceCollection container, bool runMaintenanceTasks) {
-            if (!String.IsNullOrEmpty(Settings.Current.ExceptionlessApiKey) && !String.IsNullOrEmpty(Settings.Current.ExceptionlessServerUrl)) {
+        public static void RegisterServices(IServiceProvider serviceProvider, IServiceCollection services, AppOptions appOptions, bool runMaintenanceTasks) {
```
ejsmith: I don't love that we are passing in IServiceProvider here instead of building it in here where we are mostly using it. I see why you did it because of the job bootstrapper needing it, but I'm wondering if we could refactor that some so that logic moves in here.

src/Exceptionless.Insulation/Bootstrapper.cs
```diff
@@ -26,162 +26,188 @@
...
+                services.AddSingleton<ExceptionlessClient>(client);
             }
 
-            if (!String.IsNullOrEmpty(Settings.Current.GoogleGeocodingApiKey))
-                container.ReplaceSingleton<IGeocodeService>(s => new GoogleGeocodeService(Settings.Current.GoogleGeocodingApiKey));
+            if (!String.IsNullOrEmpty(appOptions.GoogleGeocodingApiKey))
+                services.ReplaceSingleton<IGeocodeService>(s => new GoogleGeocodeService(appOptions.GoogleGeocodingApiKey));
 
-            RegisterMetricsReporting(container);
+            RegisterCache(services, serviceProvider.GetRequiredService<IOptions<CacheOptions>>().Value, appOptions);
```
ejsmith: These look really good.

src/Exceptionless.Web/Controllers/AdminController.cs
```diff
@@ -112,7 +169,7 @@ public class AdminController : ExceptionlessApiController {
         public async Task<IActionResult> RunJobAsync(string name) {
             switch (name.ToLowerInvariant()) {
                 case "indexes":
-                    if (!Settings.Current.DisableIndexConfiguration)
+                    if (!AppOptions.Current.DisableIndexConfiguration)
```
ejsmith: Why are we using AppOptions here when we have access to _appOptions?

niemyjski: overlooked

src/Exceptionless.Core/Jobs/Elastic/CleanupSnapshotJob.cs
niemyjski: We have reload=true so we could and it would be nice to support it.

src/Exceptionless.Core/Jobs/EventPostsJob.cs
niemyjski: It's what we were doing before. I think it was for a buffer / container overhead.

src/Exceptionless.Core/Repositories/Configuration/ExceptionlessElasticConfiguration.cs
niemyjski: ok

src/Exceptionless.Core/Repositories/Configuration/Indexes/OrganizationIndex/OrganizationIndex.cs
niemyjski: but we have app scope and that affects other things

src/Exceptionless.Core/Jobs/Elastic/CleanupSnapshotJob.cs
ejsmith: It would be a bit scary for some of these things to change on the fly. Seems like we would need to look at all the usages and see which options are accessed many times while the app is running and determine if its a good or bad idea for them to be changing on the fly.

src/Exceptionless.Core/Jobs/EventPostsJob.cs
ejsmith: Any idea why? Seems really freakin' strange to me. I could see us maybe multiplying by 1024 because we want the option value to be in megabytes or something, but that's not what we are doing.

src/Exceptionless.Core/Repositories/Configuration/Indexes/OrganizationIndex/OrganizationIndex.cs
ejsmith: Ok, what's your point? An option can be in more than one options class.

ejsmith:
> I think the SmptEncryption can be determined by the SmtpPort.

I like doing those kind of implied configurations. But the options class itself should be explicit properties. It's just the logic that populates those properties that would do smart things like infer an option based on something like the port number.

src/Exceptionless.Core/Jobs/EventPostsJob.cs
niemyjski: It was for the container overhead. I can add 1024 to it.

src/Exceptionless.Core/Repositories/Configuration/Indexes/OrganizationIndex/OrganizationIndex.cs
niemyjski: If we do that it seems like we should add scope to storage as well

src/Exceptionless.Core/Repositories/EventRepository.cs
niemyjski: To me retention has nothing to do with elastic search

ejsmith: It’s max days that we store events in ES as a whole system. It absolutely has to do with ES config. And we are also trying to get rid of static config access as well as try to simplify things.

niemyjski: it's also about storage too

src/Exceptionless.Core/Repositories/Configuration/Indexes/OrganizationIndex/OrganizationIndex.cs
ejsmith: That’s fine. Again, we are trying to simplify. I don’t like having multiple options classes in a single class and I want to simplify things as much as possible.

src/Exceptionless.Core/Repositories/EventRepository.cs
ejsmith: We are in the repos, we have access to ES config passed into the ctor. It seems pretty straight forward to me that we would want to not be using static config when we can easily use options from the ES config that make perfect sense being in there.

tests/Exceptionless.Tests/appsettings.yml
```diff
@@ -1,7 +1,11 @@
 ﻿---
 ConnectionStrings:
-  Redis.off: localhost,abortConnect=false
-  Elasticsearch.off: http://localhost:9200/
+  elasticsearch.off: server=http://localhost:9200
```
ejsmith: I want to be able to just do a redis connectionstring and have it be used for all of the services that use redis if there isn't any overriding configuration for that specific service. For example, in our Azure environment, we would just configure Elasticsearch and Redis connection strings and then also provide a storage connection string that says to use Azure Blob Storage and then an additional queue connection string that say to use Azure Service Bus for queues. Everything else would default to using the redis connection string like the cache and messagebus. If we wanted to use Redis for queues as well then you could just remove the overriding queue connection string and it would fall back to using Redis. If no Redis connection string was defined then everything would fallback to using in memory. Basically, the idea is that you configure services that are available in the environment and we derive the best configuration from them unless you specifically override a service config and tell us exactly what to do.

tests/Exceptionless.Tests/appsettings.yml
```diff
@@ -1,7 +1,11 @@
...
+  cache.off: provider=redis;localhost,abortConnect=false
+  messagebus.off: provider=redis;localhost,abortConnect=false
```
ejsmith: This is not a valid connection string format. It would have to be `provider=redis;server="localhost,abortConnect=false"`

src/Exceptionless.Core/Configuration/QueueOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+            _configuration = configuration;
+        }
+
+        public void Configure(QueueOptions options) {
+            options.Scope = _configuration.GetValue<string>(nameof(options.Scope), String.Empty);
+            options.ScopePrefix = !String.IsNullOrEmpty(options.Scope) ? options.Scope + "-" : String.Empty;
+
+            string cs = _configuration.GetConnectionString("queue");
+            options.Data = cs.ParseConnectionString();
+            options.Provider = options.Data.GetString(nameof(options.Provider).ToLowerInvariant());
```
ejsmith: I want to rely on the IConfiguration system to get these settings and for it to provide the conversion helpers to coerce a string value into whatever type we need. It already has all of that functionality. So what we need to do is to parse the queue connection string into a dictionary like we are doing, but then to add those key/value pairs to the IConfiguration values in a sub-section called queue. Does that make sense?

tests/Exceptionless.Tests/appsettings.yml
niemyjski: I agree and disagree. I think that we could do provider=xyz and then look for a connection string with that name and merge that connection string into the settings. I don't want a lot of configuration but I also want configuration to be explicit and not magic. I don't want to have to look at all the other options to see what's using Redis...

src/Exceptionless.Core/Configuration/QueueOptions.cs
niemyjski: yeah, that makes sense but where does that happen? This is in configure and I don't think we can add to configuration at that time. You'd basically have to say add provider and these Configure options function is called multiple times.

src/Exceptionless.Insulation/Bootstrapper.cs
```diff
@@ -26,162 +26,188 @@
...
+                container.ReplaceSingleton(s => CreateAzureStorageQueue<EventPost>(s, options, retries: 1));
+                container.ReplaceSingleton(s => CreateAzureStorageQueue<EventUserDescription>(s, options));
+                container.ReplaceSingleton(s => CreateAzureStorageQueue<EventNotificationWorkItem>(s, options));
+                container.ReplaceSingleton(s => CreateAzureStorageQueue<WebHookNotification>(s, options));
+                container.ReplaceSingleton(s => CreateAzureStorageQueue<MailMessage>(s, options));
+                container.ReplaceSingleton(s => CreateAzureStorageQueue<WorkItemData>(s, options, workItemTimeout: TimeSpan.FromHours(1)));
+            }
+        }
+
+        private static void RegisterStorage(IServiceCollection container, StorageOptions options) {
```
DaniilSokolyuk: Can you add support aws s3 (from foundatio.aws) please?

niemyjski: Sure :)

src/Exceptionless.Core/Configuration/QueueOptions.cs
niemyjski: I just made some changes for this, so if you do `provider=redis` it will pull `config.getconnectionstring(provider)` and then append all those option pairs.

src/Exceptionless.Insulation/Bootstrapper.cs
niemyjski: Added support for sqs and s3. it would be nice to have cloud watch support but that's a bit more than I want to chew off right now and could always be a future pr ;)

tests/Exceptionless.Tests/appsettings.yml
niemyjski: fixed

PR #396, opened 2018-11-16
Add new web hook types，Error and LogError。

justmine66:
When I configured webhook in the production environment, the following requirements could not be satisfied：
1. Could not receive notification error messages in real time, only the first error notification can be received.
2. Unable to notify error message in log.
That's what we care most about，I think most people have the same needs. so we decided to submit the code to master, work together to perfect this great logging component.

finally angular ui to config:
image


niemyjski:
Thanks for the PR. This change would log every error or log error occurrence and we have been against this in the past as it would be extremely noisy and alerts are no longer important. I could see us adding support for Critical Log message using the existing types (or would that be  confusing as there is error and critical error). Do you have any thoughts @ejsmith

ejsmith:
I am not sure about the ability to notify on every event, but maybe we could do something that notifies you on every event but throttles it. So that you wouldn’t get more than 1 notification every 15 mins or so.

As far as the log event type, I don't really want to call out the log type specifically. The event types are open ended, but there are 4 primary types right now. I'd like to see us allow changing notification settings for each event type. @justmine66 could you make it so that we can control notification settings per event type?

justmine66:
@ejsmith In fact, these two types(Error logError) meet the needs of production environment, at least for us. For example, an order payment failed, and its event type was logerror etc, each company records logs differently.

ejsmith:
That’s exactly right. Which is why I’d like to have our notification settings be more flexible and handle all event types.

niemyjski:
@justmine66 Can you please update this to support all event types (etc.. feature, error, log....) as well as maybe add some special cased ones for logs (@ejsmith)

niemyjski:
We really appreciate the pr. There are conflicts and no response to previous questions. I'm closing this as we want to support all event types.

PR #405, opened 2019-03-08
Enable running UI in Web

Phyxion:
Two changes are required.
1) .less files need to be served.
2) gravatar needs to be allowed for user images (CSP).

### DIFF
--- a/src/Exceptionless.Web/Startup.cs
+++ b/src/Exceptionless.Web/Startup.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.IO;
 using System.Security.Claims;
 using Exceptionless.Core;
@@ -20,6 +20,7 @@ using Joonasw.AspNetCore.SecurityHeaders;
 using System.Collections.Generic;
 using Foundatio.Hosting.Startup;
 using Microsoft.AspNetCore.Diagnostics.HealthChecks;
+using Microsoft.AspNetCore.StaticFiles;
 using Microsoft.Extensions.Options;
 
 namespace Exceptionless.Web {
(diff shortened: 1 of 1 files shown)


Phyxion:
The build failure is unrelated, it builds fine.

ejsmith:
We compile the UI project down to static HTML so I'm not sure why you would need to serve less files. Are you sure that you are doing a production build of the UI site?

Phyxion:
Yeah, I have it built to one js file, but for testing this is easier. If you don't want this, I understand, but Gravatar should at least be added :)

ejsmith:
No, I was just curious. I think this looks good. I'd like to create a docker image that contains both the UI and API like this.

Phyxion:
OK, great, thanks. I initially used the "full package" for testing until it worked then compiled it down (which also means the `@@version` is gone). This is just a little easier than having to compile the whole thing every time a change is made (although there wasn't much required in the end). I have it running OK now with these changes and it also runs better than the previous version from 2017 👍

ejsmith:
Also, just an FYI, we are finalizing a massive effort to port the UI forward to latest Angular. That code is here: https://github.com/exceptionless/Exceptionless.UI/tree/feature/angular6

Hope to get that deployed very soon.

Phyxion:
Nice, I've not seen that yet.

Little feedback (if you want it already), without --prod I have `@@version` shown in UI and ng build --prod build fails here:

> ERROR in Error during template compile of 'AppModule'
>   Function calls are not supported in decorators but 'IntercomModule' was called.

The new graphs (and options) are nice :) but do seem to render rather slow (noticeable slower than current UI). Especially when switching to another screen the animation seems to slow it down considerably.

Back to old version for now but can do another try if it would help 👍

ejsmith:
@Phyxion I would love some help with QA on the new site. Any chance you go through it and make a list of any issues that you are seeing like the slow charts?

Phyxion:
Left some feedback on the PR there :) Let's continue there if anything needs clarification.

PR #406, opened 2019-03-11
Fix NRE in DailySummaryJob

Phyxion:
These two somehow returned null here.


Phyxion:
I'm not sure @niemyjski, it was the first (and only) exception I've gotten after upgrading from the old Exceptionless. I have all summary reports disabled (and no email addresses specified for sending out emails). Could that be it?

niemyjski:
It shouldn't be the issue, if you are specifying the age it should be returned empty... Thanks again for the pr.

PR #412, opened 2019-04-06
Added ability to disable/enable webhooks and better handle timeouts.

niemyjski:


niemyjski:
I debugged it a bit and think I have it all handled all scenarios. The only issue I know with the code and I’m fine with it is your consecutive errors increment could go longer than three days, because you are incrementing the timeout by 3 days every time you encounter an error… so that key could last longer than the first attempt
We are also potentially doing an extra cache call to set last attempt if consecutive errors is greater than 10 but I’m also fine with that as it reads better

Issue #417, opened 2019-04-30
BasicAuthentication for elasticsearch in .net core version?


witskeeper:
BasicAuthentication for elasticsearch in .net core version?

niemyjski:
I'm not sure what you are asking, you want to authenticate to elastic search using basic auth? I'm not sure if the nest client supports this: https://github.com/elastic/elasticsearch-net/issues/2505 If it does, please open a pr for this, it would be easy to add support for it via the elastic options configuration class.

niemyjski:
Thanks for the pr!

PR #423, opened 2019-05-14
add BasicAuthentication for elasticsearch

witskeeper:
for issue #417 BasicAuthentication for elasticsearch in .net core version?

### DIFF
--- a/src/Exceptionless.Core/Configuration/ElasticsearchOptions.cs
+++ b/src/Exceptionless.Core/Configuration/ElasticsearchOptions.cs
@@ -17,6 +17,9 @@ namespace Exceptionless.Core.Configuration {
 
         public bool EnableSnapshotJobs { get; set; }
         public bool DisableIndexConfiguration { get; set; }
+
+        public string Password { get; internal set; }
+        public string UserName { get; internal set; }
     }
 
     public class ConfigureElasticsearchOptions : IConfigureOptions<ElasticsearchOptions> {
@@ -50,6 +53,10 @@ namespace Exceptionless.Core.Configuration {
             options.FieldsLimit = fieldsLimit > 0 ? fieldsLimit : 1500;
 
             options.EnableMapperSizePlugin = pairs.GetValueOrDefault("enable-size-plugin", _appOptions.Value.AppMode != AppMode.Development);
+
+            options.UserName = pairs.GetString("username", "");
+
+            options.Password = pairs.GetString("password", "");
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/Exceptionless.Web/appsettings.Production.yml
```diff
@@ -1,6 +1,6 @@
 ﻿---
 ConnectionStrings:
-#  Elasticsearch: server=http://localhost:9200
```
niemyjski: Can we the changes to the connection strings.

src/Exceptionless.Job/appsettings.Production.yml
```diff
@@ -1,6 +1,6 @@
 ---
 ConnectionStrings:
-#  Elasticsearch: server=http://localhost:9200
```
niemyjski: Can we the changes to the connection strings.

src/Exceptionless.Core/Repositories/Configuration/ExceptionlessElasticConfiguration.cs
```diff
@@ -60,7 +60,9 @@ public sealed class ExceptionlessElasticConfiguration : ElasticConfiguration, IS
             ConfigureSettings(settings);
             foreach (var index in Indexes)
                 index.ConfigureSettings(settings);
-
+            if (!string.IsNullOrEmpty(Options?.UserName)) {
```
niemyjski: We should change this to as we shouldn't need to null check it and we should verify both are set.

```cs

if (!String.IsNullOrEmpty(_options.UserName) && !String.IsNullOrEmpty(_options.Password))
    settings.BasicAuthentication(_options.UserName, _options.Password);
```

src/Exceptionless.Core/Configuration/ElasticsearchOptions.cs
```diff
@@ -50,6 +53,10 @@ public class ConfigureElasticsearchOptions : IConfigureOptions<ElasticsearchOpti
             options.FieldsLimit = fieldsLimit > 0 ? fieldsLimit : 1500;
 
             options.EnableMapperSizePlugin = pairs.GetValueOrDefault("enable-size-plugin", _appOptions.Value.AppMode != AppMode.Development);
+
+            options.UserName = pairs.GetString("username", "");
+
+            options.Password = pairs.GetString("password", "");
```
niemyjski: We shouldn't need the second parameter as it returns String.Empty if the key isn't set.

src/Exceptionless.Core/Repositories/Configuration/ExceptionlessElasticConfiguration.cs
witskeeper: I think what you mean is this or not ？
```
if (!String.IsNullOrEmpty(_options.Value.UserName) && !String.IsNullOrEmpty(_options.Value.Password))
    settings.BasicAuthentication(_options.Value.UserName, _options.Value.Password);
```

witskeeper:
i will submit pr again

PR #439, opened 2019-07-31
Merging the UI into the web project

ejsmith:
- [x] Serve the SPA app files from the Web project
- [x] Automatically start the SPA site (`npm run serve`) when running the web project
- [x] Make it so that doing `dotnet publish` will build the `ClientApp` and add the files to the output
- [x] Clean up some old dependencies like puppeteer
- [x] Automatically use same SSL cert from the ASPNETCORE app
- [x] Automatically pickup host, port and ssl settings from the ASPNETCORE app
- [x] Proxy to API so requests can be run on a single domain and doesn't require CORS
- [x] Single web docker image. Get rid of UI docker image
- [x] Fix issue with fallback routing not working. Should serve SPA index page when a request comes in that doesn't match an endpoint
- [x] Make it so that when `BaseURL` is empty, it calls back to same site
- [x] Remove client version from bottom of UI. We will have a single version now
- [x] Update README to explain how to work on the front end without running the back end (set API url)
- [x] Update README to explain how to work on just the back end without running the front end (change to the Exceptionless API launch profile / remove ASPNETCORE_HOSTINGSTARTUPASSEMBLIES)
- [x] Update k8s / prod to remove UI service and use combined Web / UI instead


packages/.gitignore
```diff
@@ -0,0 +1 @@
+# place holder to make directory included when we don't have any custom nuget packages
```
niemyjski, line 1: Keeping this file here, will cause an extra folder in root to be displayed.

src/Exceptionless.Web/ClientApp/app.config.js
```diff
@@ -0,0 +1,19 @@
+(function () {
+  'use strict';
+
+  angular.module('app.config', [])
+    .constant('BASE_URL', '')
```
niemyjski, line 5: running locally or against prod, this is no longer set and causes web sockets and project configure page not to work as expected.

niemyjski (approved):
Looks great, just some minor client side issues I left comments on.

packages/.gitignore
ejsmith: LOL, that is literally what the comment says it's for that you commented on. We have had custom packages multiple times in the past, if I remove this, I need to remove the custom feed in the nuget.config and remove the folder from being copied in the dockerfile and then if we ever need custom packages again, I'll have to remember to put all those things back.

src/Exceptionless.Web/ClientApp/app.config.js
ejsmith: Updated

PR #445, opened 2019-09-06
#355 - Ability to disable an api key.

### LINKED ISSUE #355: Ability to disable the api key
Would be nice to have a disabled flag on the api key so you could enable it or disable it temporarily. If your organization is suspended we could save some work right off the bat by handling it in the auth controller.

niemyjski:
Don't merge this until we've had more testing and discussion. Fixes #355

### DIFF
--- a/src/Exceptionless.Core/Utility/SampleDataService.cs
+++ b/src/Exceptionless.Core/Utility/SampleDataService.cs
@@ -7,6 +7,7 @@ using Exceptionless.Core.Repositories;
 using Exceptionless.Core.Models;
 using Foundatio.Repositories;
 using Foundatio.Utility;
+using Microsoft.Extensions.Logging;
 
 namespace Exceptionless.Core.Utility {
     public class SampleDataService {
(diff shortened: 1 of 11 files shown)


src/Exceptionless.Core/Utility/SampleDataService.cs
```diff
@@ -65,8 +77,12 @@ public class SampleDataService {
             _billingManager.ApplyBillingPlan(organization, _billingPlans.UnlimitedPlan, user);
             organization = await _organizationRepository.AddAsync(organization, o => o.Cache()).AnyContext();
 
-            var project = new Project { Id = TEST_PROJECT_ID, Name = "Disintegrating Pistol", OrganizationId = organization.Id };
-            project.NextSummaryEndOfDayTicks = SystemClock.UtcNow.Date.AddDays(1).AddHours(1).Ticks;
+            var project = new Project {
+                Id = TEST_PROJECT_ID,
```
niemyjski: @ejsmith is there any reason we used a second const id here instead of just using TestConstants?

niemyjski:
@ejsmith Please review and merge if all checks out :)

Issue #332, opened 2017-11-16
Elastic 7.0 support and Index Improvements


niemyjski:
#While we add support for [elasticsearch 6](https://github.com/exceptionless/Exceptionless/issues/331#issuecomment-344604049) we should also fix all known index issues as outlined here:

- [x] Convert Foundatio Parsers
- [x] Convert Foundatio Repositories
- [x] Searching custom extended data field of type string not returning results [#325]
- [x] Look into and see if any work needs to be done on OS Level Filtering [#324]
- [x] Look into indexing url segments or at least the domain name [#323]
- [x] Look into analyzer/tokenizer to see if we can remove punctuations [#301]
- [x] Look into wildcard search settings for optimal perf.
- [x] Test data migration in dev environment
- [x] Full round of testing in dev against migrated data
- [x] Figure out why the charts don't have as many data points as they should have
- [x] Make new storage queues? Possibly stop doing post archives
- [x] Use prod as app scope
- [x] Migrate data to prod
- [x] Setup and verify snapshots are running
- [x] Stop production application and let queues drain
- [x] Upgrade application to ES7
- [x] Do incremental reindex to get data current from old ES
- [x] Start application
- [x] Delete all old resources (ES5, etc)
- [x] Profit

niemyjski:
Looks like there may be an issue with the query parser for issue #325, the query seems to be generated incorrectly some of the time.

einarbey:
Hi 

i have a new setup of elasticsearch and exceptionless and i get this error 

Error creating the index stacks-v1: Original: (400 - WebException) The remote server returned an error: (400) Bad Request.
Server: (400) Type: mapper_parsing_exception Reason: "Failed to parse mapping [stacks]: [include_in_all] is not allowed for indices created on or after version 6.0.0 as [_all] is deprecated. As a replacement, you can use an [copy_to] on mapping fields to create your own catch all field." CausedBy: "Type: mapper_parsing_exception Reason: "[include_in_all] is not allowed for indices created on or after version 6.0.0 as [_all] is deprecated. As a replacement, you can use an [copy_to] on mapping fields to create your own catch all field.""

i am running elasticsearch 6.x how can i manually create this index with out having to install elastic 5.x ?

niemyjski:
We do not support elasticsearch 6, you'll have to use elasticsearch 5.

woshilangdanger:
When can elasticsearch 6.0 be supported?

niemyjski:
[...]

PR #452, opened 2019-10-01
Elasticsearch 7 support

niemyjski:
This tracks #332


src/Exceptionless.Core/Serialization/ElasticJsonNetSerializer.cs
```diff
@@ -0,0 +1,38 @@
+using System;
+using System.Linq;
+using Elasticsearch.Net;
+using Nest;
+using Nest.JsonNetSerializer;
+using Newtonsoft.Json;
+
+namespace Exceptionless.Core.Serialization {
+    public class ElasticJsonNetSerializer : JsonNetSerializer {
```
niemyjski: @ejsmith we should be using all of our serializer settings

PR #735, opened 2020-10-02
Add KnownProxies configuration

### LINKED ISSUE #733: Add KnownProxies configuration option
See https://github.com/exceptionless/Exceptionless/issues/658#issuecomment-700669019

PhyxionNL:
Fixes #733.

### DIFF
--- a/src/Exceptionless.Core/Configuration/AppOptions.cs
+++ b/src/Exceptionless.Core/Configuration/AppOptions.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Net;
 using Exceptionless.Core.Configuration;
 using Exceptionless.Core.Extensions;
 using Microsoft.Extensions.Configuration;
--- a/src/Exceptionless.Core/Configuration/AppOptions.cs
+++ b/src/Exceptionless.Core/Configuration/AppOptions.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Net;
 using Exceptionless.Core.Configuration;
 using Exceptionless.Core.Extensions;
 using Microsoft.Extensions.Configuration;
@@ -45,6 +46,7 @@ namespace Exceptionless.Core {
         public bool EventSubmissionDisabled { get; internal set; }
 
         internal List<string> DisabledPipelineActions { get; set; }
+
         internal List<string> DisabledPlugins { get; set; }
 
         /// <summary>
@@ -53,6 +55,8 @@ namespace Exceptionless.Core {
         public long MaximumEventPostSize { get; internal set; }
 
         public int MaximumRetentionDays { get; internal set; }
+
+        public List<string> KnownProxies { get; internal set; }
 
         public bool EnableRepositoryNotifications { get; internal set; }
 
@@ -109,6 +113,8 @@ namespace Exceptionless.Core {
 
             options.EnableRepositoryNotifications = config.GetValue(nameof(options.EnableRepositoryNotifications), true);
             options.EnableWebSockets = config.GetValue(nameof(options.EnableWebSockets), true);
+            
+            options.KnownProxies = config.GetValueList(nameof(KnownProxies));
 
             try {
                 var versionInfo = FileVersionInfo.GetVersionInfo(typeof(AppOptions).Assembly.Location);
--- a/src/Exceptionless.Core/Configuration/AppOptions.cs
+++ b/src/Exceptionless.Core/Configuration/AppOptions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using Exceptionless.Core.Configuration;
@@ -45,7 +45,6 @@ namespace Exceptionless.Core {
         public bool EventSubmissionDisabled { get; internal set; }
 
         internal List<string> DisabledPipelineActions { get; set; }
-
         internal List<string> DisabledPlugins { get; set; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


PhyxionNL:
Squash and merge.

PhyxionNL:
Failure is unrelated.

niemyjski:
Thanks for the PR!!! I wonder if our ordering for forwarded headers is in the correct spot too https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-3.1#forwarded-headers-middleware-order Also from the discussion here I wonder if we also need to have KnownNetworks (https://github.com/dotnet/AspNetCore.Docs/issues/2384) Maybe not? All the docs only mention knownproxies (https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/linux-nginx?view=aspnetcore-3.1)  I'm curious what was the scenarios/setup you tested this with (I think you mentioned docker-compose earlier)? Also this is interesting: https://github.com/dotnet/AspNetCore.Docs/issues/12349 and https://soapfault.com/2020/02/24/asp-net-core-reverse-proxy-and-x-forwarded-headers/ cc @ejsmith

PhyxionNL:
I think the location is fine, if it even matters. I've created a reverse proxy in IIS and the `KnownProxies` must be set, otherwise .NET Core doesn't know which IP is the proxy.

`KnownNetworks` is only required to limit where the forwarded headers can come from.

ejsmith:
Thank you very much for helping us with this issue!

I noticed on [this blog post](https://devblogs.microsoft.com/aspnet/forwarded-headers-middleware-updates-in-net-core-3-0-preview-6/) they are saying to clear the `KnownProxies` list. I wonder if that would be a better way for Exceptionless to handle this issue. I think it can be somewhat of a security concern if you are using those things for any sort of access rules. In Exceptionless I believe we are simply using them to populate information about the events. So seems like it would be OK and it would make it work in all scenarios without configuration. Thoughts?

PhyxionNL:
Clearing them doesn't work with IIS, it'll need to be specified explicitly. It also says that ("To forward the scheme from the proxy in **non-IIS** scenarios"). Normally only the IPv6 loopback is added. Basically with this change it stays as-is but enables Exceptionless to be reverse proxied through IIS 😊

ejsmith:
Here is a docs page about it too: https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-3.1#forward-the-scheme-for-linux-and-non-iis-reverse-proxies

ejsmith:
> Clearing them doesn't work with IIS, it'll need to be specified explicitly. It also says that ("To forward the scheme from the proxy in **non-IIS** scenarios"). Normally only the IPv6 loopback is added. Basically with this change it stays as-is but enables Exceptionless to be reverse proxied through IIS 😊

Hmm... they seemed to indicate that it would work with IIS automatically.

PhyxionNL:
I've tried that already, but it doesn't work for me unless KnownProxies is correctly specified.

ejsmith:
Did you try adding `UseIISIntegration`?

ejsmith:
Actually, that should be on already as `ConfigureWebHostDefaults` appears to do it.

PhyxionNL:
Yeah, I've also tried that, but it doesn't help because it's not running through IIS but inside docker. If I run Exceptionless straight through IIS (i.e. hosted) it's working correctly. The block of code from https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-3.1#forward-the-scheme-for-linux-and-non-iis-reverse-proxies doesn't work, I've tried it (and variations of it) several times it it's always ::ffff:172.18.0.1. This host needs to be added to `KnownProxies`, otherwise the wrong IP gets reported. This is exactly where `KnownProxies` is for.

`The request's original remote IP must match an entry in the KnownProxies or KnownNetworks lists before forwarded headers are processed. This limits header spoofing by not accepting forwarders from untrusted proxies.`

ejsmith:
Ok, hmm... so my understanding is that the KnownProxies list is basically white listing entries, but if the list is empty it should work always if those headers are present. I'm just trying to understand this because ideally I don't want to have to add an extra configuration step.

Looking at the code here: https://github.com/dotnet/aspnetcore/blob/master/src/Middleware/HttpOverrides/src/ForwardedHeadersMiddleware.cs#L223

It seems like from reading that code if both of those collections are empty then it bypasses any known hosts checks and if the headers are present it should use them. But you are saying that you set both collections to empty and it did not apply your incoming `X-Forwarded` headers, correct?

PhyxionNL:
No, this must be set for forwarding headers to be accepted. It's only working for loopback. See above. 

For what it's worth, it's also possible to use `KnownNetworks`, e.g. `options.KnownNetworks.Add(new IPNetwork(IPAddress.Parse("172.18.0.0"), 16));` But this format isn't very friendly as it's not CIDR notation. I could add this also and make it something like KnownNetworks: strings in format IP-PrefixLength

PhyxionNL:
Update: clearing all Known... properties seems to work as well, let me try some more.

For what it's worth, it's also possible to use `KnownNetworks`, e.g. `options.KnownNetworks.Add(new IPNetwork(IPAddress.Parse("172.18.0.0"), 16));` But this format isn't very friendly as it's not CIDR notation. I could add this also and make it something like KnownNetworks: strings in format IP-PrefixLength

ejsmith:
:-) I was going to say... I read that code and I don't see how it would be checking if both are empty. I feel like in general they need to be careful with setting these things because people might be making security decisions based on their values. But in our case we are not and in general if we are able to just use these always and not have to have a config option for this, that is what I would prefer. Can you do some testing and let me know what you find out? Thank you!!

PhyxionNL:
That indeed seems to work as well. I'll update to the code to clear them all by default, that seems to be enough 😊

ejsmith:
Awesome! Thank you @PhyxionNL !!

PhyxionNL:
https://github.com/exceptionless/Exceptionless/pull/736

niemyjski:
Thanks!

Issue #750, opened 2020-10-21
[Selhosted] Exceptionless sending mail to fictive mail


Nonobis:
Hello,

Every night exceptionless send report mail for fictive project to fictive address. I receive bad sending report from google.

- It's marked as free plan with random project
- Target : free@test.com, test@test.com

Environnement :
i am on docker, EX_AppMode is on Production, mail settings are configured by EX_ConnectionStrings__Email

=> Email are defined in SampleDataService

Thanks

niemyjski:
We're really sorry about that. I'm not sure we considered this happening as we thought it would just be dev mode and not promoted to production mode. If production mode is specified we won't create sample data which will have the test user. You could log in as the test and free  users and delete them in the mean time. I'm going to mark that email address as not verified so a user will have to verify it to get emails sent.

niemyjski:
I didn't add a migration for this to find them and mark them as unverified but we'd accept a pr for this if you'd like to submit one.

Nonobis:
@niemyjski Thanks, i will see if i got time to make a PR. Because i receive many mail.... ^^

niemyjski:
You could log in and delete those accounts or just remove them via elastic

PR #751, opened 2020-10-21
When AppMode is in 'Production' set IsEmailAddressVerified and IsActi…

Nonobis:
…ve to false when creating test user.
PR for solving #750 

Can't test here but should be ok. 
Tell me if you want that i refactorise or change code.


Nonobis:
sorry, i will fix it...

niemyjski:
Thanks for the pr, I already solved this in the commit that closed #750. I was referring to a migration that updated existing environments automatically via a migration: https://github.com/exceptionless/Exceptionless/tree/master/src/Exceptionless.Core/Migrations where we try and get the free and test user and mark them as not verified (https://github.com/exceptionless/Exceptionless/blob/master/src/Exceptionless.Core/Utility/SampleDataService.cs#L22-L30)

niemyjski:
No problem, thanks for your time! If I can help with anything please let me know.

PR #829, opened 2021-01-30
Web: Remove boxing and params allocations from Info logging calls

benaadams:
User `LoggerMessage.Define` to avoid allocations from boxing and params allocations from Information level logging; which occur regardless as to whether that level is enabled.


ejsmith (approved):
Looks great! I like the pattern you established here.

PR #830, opened 2021-01-31
Reduce console logging costs

benaadams:
On Windows 97% of `EventController.PostAsync` is console logging

image

Switching it to ansi mode (for Development) and no colors for Staging/Production reduces that to 75% https://github.com/serilog/serilog-sinks-console/issues/58

image

Changing it to be >= `Warning` for Staging/Production removes the cost entirely (assuming no errors 😉)

image


benaadams:
Will need extra checking that its behaving in Staging as I assume the tests aren't running with Staging/Production's settings

ejsmith (approved):
Looks good. Thanks!

PR #837, opened 2021-02-13
Status409Conflict -> Status429TooManyRequests

benaadams:
To indicate rate limiting; could also be `Status402PaymentRequired` to signal upgrade required?

When load testing `Status409Conflict` confused me as it suggested duplicate request issues.


ejsmith:
Agreed. Thanks!

ejsmith (approved):

PR #838, opened 2021-02-13
Core: Remove boxing and params allocations from logging calls

benaadams:
User `LoggerMessage.Define` to avoid allocations from boxing and params allocations from logging; which occur regardless as to whether that level is enabled.


ejsmith (approved):
Thank you! Are you seeing that the log messages not using pre-compiled templates are one of the major issues or more that changing these are an easy win?

PR #839, opened 2021-02-13
Job: Use nameof()

benaadams:
Minor change 😉


src/Exceptionless.Job/JobRunnerOptions.cs
```diff
@@ -78,7 +78,7 @@ public class JobRunnerOptions {
         public bool CloseInactiveSessions { get; }
         public bool DailySummary { get; }
         public bool DataMigration { get; }
-        public bool DownloadGeoipDatabase { get; }
+        public bool DownloadGeoIPDatabase { get; }
```
benaadams, line 81: Changed to match the `JobName` string that was being used previously so it would have same output when using `nameof`

ejsmith: This should be fine. Running jobs by name isn't case sensitive.

ejsmith (approved):
LGTM. Thank you!

PR #941, opened 2021-08-09
Cleanup open stacks 60 days older than the retention period that have no references.

niemyjski:


src/Exceptionless.Core/Repositories/StackRepository.cs
```diff
@@ -26,6 +26,14 @@ public StackRepository(ExceptionlessElasticConfiguration configuration, IValidat
             return FindAsync(q => q.ElasticFilter(Query<Stack>.DateRange(d => d.Field(f => f.SnoozeUntilUtc).LessThanOrEquals(utcNow))), options);
         }
 
+        public Task<FindResults<Stack>> GetStacksForCleanupAsync(DateTime utcNow, CommandOptionsDescriptor<Stack> options = null) {
```
ejsmith, line 29: This should be called cutoff

src/Exceptionless.Core/Repositories/StackRepository.cs
```diff
@@ -26,6 +26,14 @@ public StackRepository(ExceptionlessElasticConfiguration configuration, IValidat
...
+            return FindAsync(q => q
+                .ElasticFilter(Query<Stack>.DateRange(d => d.Field(f => f.LastOccurrence).LessThanOrEquals(utcNow)))
+                .FieldEquals(f => f.Status, StackStatus.Open)
+                .FieldEmpty(f => f.References)
```
ejsmith, line 33: Is this the only thing we need to check for custom data put on the stack?

ejsmith (changes requested):

src/Exceptionless.Core/Repositories/StackRepository.cs
niemyjski: The only custom data we have on a stack is reference links

PR #1025, opened 2021-12-28
Add safety checks to prevent negative usage.

niemyjski:


src/Exceptionless.Web/Utility/Handlers/OverageMiddleware.cs
```diff
@@ -87,14 +87,15 @@ public sealed class OverageMiddleware {
 
         var organization = await organizationTask;
         var project = await projectTask;
-        bool overLimit = await _usageService.IncrementUsageAsync(organization, project, tooBig);
 
         // block large submissions, client should break them up or remove some of the data.
         if (tooBig) {
+            await _usageService.IncrementTooBigAsync(organization, project).AnyContext();
```
ejsmith, line 93: Is this a new counter?

src/Exceptionless.Core/Extensions/CacheClientExtensions.cs
```diff
@@ -16,10 +16,10 @@ public static class CacheClientExtensions {
...
         if (count.HasValue) {
             var incrementedValue = await client.IncrementAsync(key, value, timeToLive).AnyContext();
-            return incrementedValue;
-            if (incrementedValue < 0) {
-                return await client.IncrementAsync(key, Math.Abs(incrementedValue), timeToLive).AnyContext();
-            }
+            if (incrementedValue >= 0)
+                return incrementedValue;
+
+            return await client.IncrementAsync(key, Math.Abs(incrementedValue), timeToLive).AnyContext();
```
ejsmith, line 22: Don't love this. It feels hacky, but I guess it will do for now.

src/Exceptionless.Core/Jobs/EventPostsJob.cs
```diff
@@ -110,7 +110,9 @@ public class EventPostsJob : QueueJobBase<EventPost> {
 
             _metrics.Gauge(MetricNames.PostsUncompressedSize, payload.LongLength);
             if (uncompressedData.Length > maxEventPostSize) {
-                await Task.WhenAll(CompleteEntryAsync(entry, ep, SystemClock.UtcNow), organizationTask).AnyContext();
+                var org = await organizationTask.AnyContext();
```
ejsmith, line 113: Seems like we could end up trying to await this task multiple times, but I guess that's ok?

ejsmith (approved):

src/Exceptionless.Web/Utility/Handlers/OverageMiddleware.cs
niemyjski: It's a new utility method around increment.

src/Exceptionless.Core/Jobs/EventPostsJob.cs
niemyjski: It shouldn't be, but even if so I think that is fine.

src/Exceptionless.Core/Extensions/CacheClientExtensions.cs
niemyjski: I don't love it either, I think we can revert this change soon after any existing data is cleaned up.

PR #1112, opened 2022-06-15
Added ability to send slack webhook messages if the url ends with /slack

niemyjski:
- This allows you to send messages to discord using the slack url suffix,
- This doesn't work when promoting stacks.

I also had to base this off the usage branch due to metrics and some other minor changes

### DIFF
--- a/src/Exceptionless.Core/Plugins/WebHook/Default/005_SlackPlugin.cs
+++ b/src/Exceptionless.Core/Plugins/WebHook/Default/005_SlackPlugin.cs
@@ -16,10 +16,13 @@ public sealed class SlackPlugin : WebHookDataPluginBase {
             return Task.FromResult<object>(null);
 
         var error = ctx.Event.GetError();
-        if (error is null)
+        if (error is null) {
+            ctx.IsCancelled = true;
             return Task.FromResult<object>(null);
+        }
 
         var message = _pluginManager.GetSlackEventNotificationMessage(ctx.Event, ctx.Project, ctx.Event.IsCritical(), ctx.IsNew, ctx.IsRegression);
+        ctx.IsCancelled = message is null;
         return Task.FromResult<object>(message);
     }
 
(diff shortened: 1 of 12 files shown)


src/Exceptionless.Core/Plugins/WebHook/Default/005_SlackPlugin.cs
```diff
@@ -0,0 +1,35 @@
...
+[Priority(5)]
+public sealed class SlackPlugin : WebHookDataPluginBase {
+    private readonly FormattingPluginManager _pluginManager;
+
+    public SlackPlugin(FormattingPluginManager pluginManager, AppOptions options) : base(options) {
+        _pluginManager = pluginManager;
+    }
+
+    public override Task<object> CreateFromEventAsync(WebHookDataContext ctx) {
+        if (String.IsNullOrEmpty(ctx.WebHook.Url) || !ctx.WebHook.Url.EndsWith("/slack"))
```
ejsmith, line 15: I'm confused, if I go to discord and tell it to create a webhook URL for me, I don't get to pick the URL. So how can I make it end with `/slack`? And if the purpose of this PR is to trigger the webhook data to be formatted a certain way then I'd rather see us do that with a query string parameter. Also, we should consider automatically doing it for known endpoints like `https://discord.com/api/webhooks/`. If the URL starts with that, then we know it's a discord webhook so we could automatically trigger it to send the payload in a format that discord understands.

ejsmith: Also, I'm guessing that discord supports the slack message format which is why we are doing this?

ejsmith (commented):
I like the idea of us having the ability to add something to a webhook url to trigger a data format change. Seems like we should have a few different formats that we support and trigger it by something we can add to a webhook url and still have it work like a query string parameter. Also, I think we should sniff URLs to get smart defaults but allow overriding it with the query string parameter.

ejsmith:
I just looked at the discord docs and I see now that discord allows you to append `/slack` to get slack compatible webhook handling. I still think the checking for `/slack` at the end of the URL is a hacky way of triggering a format though. What if some other service comes along that does it differently, but they still support slack format webhooks?

PR #1258, opened 2023-02-22
Fixes #1228: Load project config from request and don't log out if project doesn't exist

### LINKED ISSUE #1228: Project identifier not being passed to Exceptionless Web app
Using the Exceptionless.AspNetCore and an example provided on your site, I set up my Web API to send events to the local Exceptionless Web app. All is working, but looking at the console logs of Exceptionless, when posting an event, I noticed the following: 
```
Project null from request doesn't match project route id 63c84998693cea2714470263
```
Is it something I am missing or is it just an information log that I shouldn't worry about?

Client set up:

appsettings.json:
```json
"Exceptionless": {
    "ApiKey": "7Rmf85EqbDksj8XAe4vSD2RBL2UYVsftmMO21goa",
    "ServerUrl": "https://localhost:7101"
}
```

niemyjski:
The project will never be set on the request by default.

### DIFF
--- a/src/Exceptionless.Web/Controllers/EventController.cs
+++ b/src/Exceptionless.Web/Controllers/EventController.cs
@@ -679,13 +679,19 @@ public class EventController : RepositoryApiController<IEventRepository, Persist
     [ConfigurationResponseFilter]
     [ProducesResponseType(StatusCodes.Status202Accepted)]
     public async Task<IActionResult> SetUserDescriptionAsync(string referenceId, UserDescription description, string projectId = null) {
+        string claimProjectId = Request.GetProjectId();
+        if (projectId is not null && claimProjectId is not null && !String.Equals(projectId, claimProjectId)) {
+            _logger.ProjectRouteDoesNotMatch(claimProjectId, projectId);
+            return NotFound();
+        }
+
         if (String.IsNullOrEmpty(referenceId))
             return NotFound();
 
         if (description == null)
             return BadRequest("Description must be specified.");
 
-        projectId ??= Request.GetDefaultProjectId();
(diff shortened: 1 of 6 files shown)


src/Exceptionless.Web/Controllers/EventController.cs
```diff
@@ -892,26 +891,19 @@ ILoggerFactory loggerFactory
...
-
-        var project = Request.GetProject();
-        if (!String.Equals(project?.Id, projectId)) {
-            _logger.ProjectRouteDoesNotMatch(project?.Id, projectId);
-
-            project = await GetProjectAsync(projectId);
-
-            // Set the project for the configuration response filter.
-            Request.SetProject(project);
-        }
```
niemyjski, line 910: Getting the project here doesn't make any sense because there is only three possible paths where the project gets set on the request (for use by the configuration header) and that's in this controller in 3 different actions...

This log message that was being outputted doesn't even make sense because it's outputted all the time when the project on the request is null. I think it was originally there to output if the request claims project id doesn't match the route project id.

Tokens can be scoped to a project id but there is no explicit tenancy check here on the project (only tenancy check to see if the user can access the associated projects organization).

ejsmith: I'm all for removing the logger message. But it seems like we should try and get the already loaded project from the request if we can. Are you saying it's not possible for that to be populated in this endpoint?

niemyjski: You are correct, it's never set before this point.

ejsmith (dismissed):
Seems good to me

niemyjski (commented):
Added tests

PR #1330, opened 2023-05-24
Use Husky.Net for code formatting on pre-commit, update build to lint

niemyjski:


ejsmith:
It seems fine, but I want to know how much `dotnet format` jacks with our code before I decide if I want to do this.

niemyjski:
@ejsmith it's making the publish failed if the ui code isn't formatted correctly. Since we don't have node installed before the build starts I put it there.

ejsmith:
I don't want that to be part of the publish step. It should be it's own step and make the build fail in a way that it's obvious what is wrong.

ejsmith:
Pulled branch and tried to do a commit and the hook failed:
image

ejsmith:
I did an npm install and now the hook worked, but it takes about 10 seconds just to do a commit now. I think prettier is running on the entire code base and not just changed files.

ejsmith:
It is actually dotnet format that is taking the full 10 seconds to run on a single .cs file. That is crazy. Apparently it is a known issue.

PR #1334, opened 2023-05-30
Fixes #1316 Preserve extended data titles

### LINKED ISSUE #1316: Preserve string sent with AddObject and not try to make them more readable by splitting the keys in the ui based on casing
I use AddObject quite a lot to send additional information with an error e.g. 

`ex.ToExceptionless().AddObject(SelectedBackupType, "SelectedBackupType").Submit() `

And what I enter as the name e.g. in this example SelectedBackupType is what I want to appear in the web interface, as in most cases I use the variable name so I can easily refer to it later.

So why does it insist on inserting spaces between words when I don't want them? it happens with most items I submit with .AddObject and although is only a "visual" annoyance, is still annoying.

Screenshot below shows other examples:

image

niemyjski:
Fixes https://github.com/exceptionless/Exceptionless/issues/1316

This change also effects anything that doesn't have a defined title:

image

# Previous
image

# New Behavior
image

### DIFF
--- a/src/Exceptionless.Web/ClientApp/app/event/event.tpl.html
+++ b/src/Exceptionless.Web/ClientApp/app/event/event.tpl.html
@@ -15,13 +15,13 @@
       </div>
       <div class="panel-body">
         <uib-tabset class="tab-container hidden-xs" active="vm.activeTabIndex">
-          <uib-tab ng-repeat="tab in vm.tabs" heading="{{::tab.title | toSpacedWords | translate}}" ng-click="vm.activateTab(tab.title)">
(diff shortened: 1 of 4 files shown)


ejsmith:
If we are going to make this change, then the tab title should match and be consistent. Are people able to have spaces in custom event data names? Could they give a friendly name themselves?

ejsmith:
Right, but can they use spaces for the name in `SetProperty` ?

niemyjski:
> Right, but can they use spaces for the name in `SetProperty` ?

Yes

PR #1385, opened 2023-08-01
Fixes #1382: Added Search Before and After stable paging

### LINKED ISSUE #1382: Get Events via API: The parameter "after" is not working
When using the REST-API to get all events, the parameter "after" is not working. 
The result is always an empty array.
I use this route `/api/v2/projects/:projectId/events `

niemyjski:


ejsmith (approved):
LGTM, lots of changes for a paging issue, but it seems good.

PR #1469, opened 2023-11-29
Change jobs to run on cron schedule

### LINKED ISSUE #1467: MaximumRetentionDays not work in self-host environment?
reference to #919, i understand that can setting EX_MaximumRetentionDays environment variable to regular cleanup the expired data.
but it does not work when i trun EX_RunJobsInProcess to false? 

is this a correct behavior for the origin design?

i use this(https://github.com/exceptionless/Exceptionless/blob/main/samples/docker-compose.yml) docker-compose.yml to start the exceptionless service.

ejsmith:
Fixes #1467


niemyjski (changes requested):
We should only run the cron job schedule when running all jobs. Otherwise when the job runs in K8S on a cron, it won't run until the schedule executes

niemyjski (approved):

PR #1496, opened 2024-02-13
Svelte Runtime Configuration

niemyjski:
1. load runtime-env.js which sets things up on the window object.
2. that is applied if set to a value otherwise uses env.js (local dev) in client.hook.js  and applys to env imported from  $env/dynamic/public
3. then looks in storage url only for base url to win out  in client.hook.js and applys to env imported from  $env/dynamic/public
4. app bootstraps as normal use $env/dynamic/public


ejsmith (approved):
Looks good to me

PR #1503, opened 2024-02-20
Dynamic Grid Filters, Updated Foundatio and OTEL

niemyjski:
image
image

- [x] Need to refactor to a builder type pattern so child controls register themselves with the parent builder component and render automatically.
- [x] This will support Organization filtering
- [x] This is just simple filtering; I think it's a good idea to only have 1 per type of filter.
- [x] User can also type a filter in via ad hoc textbox filter into part of this as well.


.github/workflows/build.yaml
```diff
@@ -121,17 +121,10 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Setup .NET Core
-        uses: actions/setup-dotnet@v4
-        with:
-          dotnet-version: 8.0.*
-          #dotnet-quality: preview
-
```
ejsmith, line 129: Why did you remove this?

niemyjski: This is for the client build. There is no reason to install .NET for it.

PR #1507, opened 2024-03-01
Discard fixed stack events that don't regress

niemyjski:
- There is an optimization to only check the regressed status on only stacks with a status of `Fixed`. As you can't regress from any other version.
- If an organization has premium features and the stack is marked fixed and doesn't regress then **those events will be discarded and won't count towards limits**.

The one downside to this implementation is that if you mark a stack as fixed, you will not see any users who are still experiencing the error to reach out to them, as those events will be discarded.

### DIFF
--- a/src/Exceptionless.Core/Pipeline/030_CheckForRegressionAction.cs
+++ b/src/Exceptionless.Core/Pipeline/030_CheckForRegressionAction.cs
@@ -23,7 +23,7 @@ public class CheckForRegressionAction : EventPipelineActionBase
     public override async Task ProcessBatchAsync(ICollection<EventContext> contexts)
     {
         var stacks = contexts
-            .Where(c => c.Stack is not null && c.Stack.Status != StackStatus.Regressed && c.Stack.DateFixed.HasValue)
+            .Where(c => c.Stack is { Status: StackStatus.Fixed, DateFixed: not null })
             .OrderBy(c => c.Event.Date)
             .GroupBy(c => c.Event.StackId);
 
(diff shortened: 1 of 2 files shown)


src/Exceptionless.Core/Pipeline/030_CheckForRegressionAction.cs
```diff
@@ -23,7 +23,7 @@ public CheckForRegressionAction(IStackRepository stackRepository, SemanticVersio
     public override async Task ProcessBatchAsync(ICollection<EventContext> contexts)
     {
         var stacks = contexts
-            .Where(c => c.Stack is not null && c.Stack.Status != StackStatus.Regressed && c.Stack.DateFixed.HasValue)
```
niemyjski, line 26: No other status should trigger a regression, so we can save a little bit of work here.

src/Exceptionless.Core/Pipeline/030_CheckForRegressionAction.cs
```diff
@@ -23,7 +23,7 @@ public CheckForRegressionAction(IStackRepository stackRepository, SemanticVersio
...
+            .Where(c => c.Stack is { Status: StackStatus.Fixed, DateFixed: not null })
```
niemyjski, line 26: No other status should trigger a regression, so we can save a little bit of work here.

src/Exceptionless.Core/Pipeline/030_CheckForRegressionAction.cs
```diff
@@ -52,7 +52,16 @@ public override async Task ProcessBatchAsync(ICollection<EventContext> contexts)
                     {
                         var version = _semanticVersionParser.Parse(versionGroup.Key, versionCache) ?? _semanticVersionParser.Default;
                         if (version < fixedInVersion)
+                        {
+                            foreach (var ctx in stackGroup.Where(s => s.Organization.HasPremiumFeatures))
```
niemyjski, line 56: I thought about adding a new pipeline action, but it would basically be a copy of this pipeline action with the same version parsing.

PR #1525, opened 2024-03-18
Account creation with token and AccountCreation false

ChaosDMG:
The api already work this way.  You can sign up if EnableAccountCreation is true or if you have a valid token linked to an organization.  Currently the frontend wont submit when you enter with a token if EnableAccountCreation is false.

Here is the relevant api logic:

    private async Task<bool> IsAccountCreationEnabledAsync(string? token)
    {
        if (_authOptions.EnableAccountCreation)
            return true;`

         if (String.IsNullOrEmpty(token))
            return false;

        var organization = await _organizationRepository.GetByInviteTokenAsync(token);
        return organization is not null;
    }


niemyjski (approved):

niemyjski:
Thanks for the PR!!!

PR #1761, opened 2024-12-01
Next: Stack Actions

niemyjski:
- Fixed issues with the grid loading and other loading indicators.
- Implemented full support for getting stack data from api and implemented all stack actions.
- Added ability to get event aggregations (fully typed)
- Fixed rendering issues with code blocks.
- Added ability for delayed rendering, useful when you only want to show indicators after x period of time
- Fixed grids saying no data when data was being loaded, skeletons now show.
- Feed through browsers time offset into api's for accurate dates.
- Fixed issues with grids constantly rerendering due to loading signals.
- Updated dependencies & shadcn components
- Tweaked readme

image
image
image
image
image


niemyjski (commented):
Self reviewed

Issue #1526, opened 2024-03-24
how to use rabbitmq as messagebus or queue?


sampsonye:
`Foundatio.Rabbitmq` use **uri-spec** as connectionstring (https://www.rabbitmq.com/docs/uri-spec)

but in Exceptionless, we can not build it like uri

![image](https://github.com/exceptionless/Exceptionless/assets/5895634/53ee391b-6ad0-4868-a066-a63dbcb490e2)

can give me an example?

sampsonye:
> We probably need to take a pass at redoing all of configuration to be more explicit, but you should be able to do this two ways.
> 
> 1. Set the MessageBus connection string to `provider=rabbitmq;CONNECTION_STRING_HERE`
> 2. Set the MessageBus connection string to ; then add rabbitmq to the connection strings app config / env section with the raw connection string`provider=rabbitmq`
> 
> cc @ejsmith
all dosn't work fine because exceptionless request connectionstring must like `key=value` patten
![image](https://github.com/exceptionless/Exceptionless/assets/5895634/c842b63c-df9d-458d-b2cb-8cec50df09a1)
`ParseConnectionString()` and `BuildConnectionString()` method working on it.

sampsonye:
![image](https://github.com/exceptionless/Exceptionless/assets/5895634/b40a12c8-6db3-4489-8287-cb9c9bd6390c)

niemyjski:
Can you paste that connection string here and also try escaping it with quotes?

niemyjski:
I was finally able to dig in and get this fixed. I've submitted a pr for the team to review.

niemyjski:
Thanks for logging this, this is fixed in a nightly build

niemyjski:
I had to revert this change as it was a pretty big breaking change to configuration. We will address this shortly.

niemyjski:
Update: PR #2273 has been expanded into a compatibility-safe infrastructure configuration change and is ready for another review: https://github.com/exceptionless/Exceptionless/pull/2273

For new configurations, RabbitMQ can now be declared directly as a named technology connection string and will automatically supply `MessageBus`:

```yaml
EX_ConnectionStrings__RabbitMQ: amqps://user:password@rabbitmq:5671/%2F
```

Raw `amqp://` and `amqps://` URIs are preserved, including encoded credentials, virtual hosts, query options, and optional wrapping quotes. The older forms remain supported too:

```yaml
[...]

PR #1784, opened 2024-12-17
Switch to using Aspire

ejsmith:


src/Exceptionless.AppHost/Extensions/ElasticsearchExtensions.cs
```diff
@@ -0,0 +1,121 @@
...
+        ArgumentNullException.ThrowIfNull(source);
+
+        return builder.WithBindMount(source, "/usr/share/elasticsearch/data");
+    }
+}
+
+internal static class ElasticsearchContainerImageTags
+{
+    public const string Registry = "docker.elastic.co";
+    public const string Image = "elasticsearch/elasticsearch";
```
niemyjski, line 118: How will this work with mapper-size plugin? Guessing that is a production only toggle.

src/Exceptionless.AppHost/Extensions/MinIoExtensions.cs
```diff
@@ -0,0 +1,138 @@
...
+        DataVolumePath = path;
+        return this;
+    }
+}
+
+internal static class MinIoContainerImageTags
+{
+    internal const string Registry = "docker.io";
+    internal const string Image = "minio/minio";
+    internal const string Tag = "latest";
```
niemyjski, line 137: would it be better to use known versions as latest will only ever be pulled once.  Makes me wonder if we could pull these versions from an env file?

src/Exceptionless.AppHost/Program.cs
```diff
@@ -0,0 +1,55 @@
+var builder = DistributedApplication.CreateBuilder(args);
+
+var elastic = builder.AddElasticsearch("Elasticsearch", port: 9200)
+    .WithImageTag("8.17.0")
```
niemyjski, line 4: We have consts for these, but I'd prefer to just have them here like this.

src/Exceptionless.Core/Configuration/MessageBusOptions.cs
```diff
@@ -22,12 +22,27 @@ public static MessageBusOptions ReadFromConfiguration(IConfiguration config, App
         options.Topic = config.GetValue<string>(nameof(options.Topic), $"{options.ScopePrefix}messages")!;
 
         string? cs = config.GetConnectionString("MessageBus");
-        options.Data = cs.ParseConnectionString();
-        options.Provider = options.Data.GetString(nameof(options.Provider));
+        if (cs != null)
```
niemyjski, line 25: Really wonder if we should just gut our provider-based connection strings only thing it does is make it more pluggable and I guess is nice for aws vs azure vs other but the complexity. Especially https://github.com/exceptionless/Exceptionless/issues/1526

tests/Exceptionless.Tests/AppWebHostFactory.cs
```diff
@@ -1,12 +1,37 @@
...
+    {
+        var options = new DistributedApplicationOptions { AssemblyName = typeof(ElasticsearchResource).Assembly.FullName, DisableDashboard = true };
+        var builder = DistributedApplication.CreateBuilder(options);
+
+        // don't use random ports for tests
+        builder.Configuration["DcpPublisher:RandomizePorts"] = "false";
+
+        builder.AddElasticsearch("Elasticsearch", port: 9200)
+            .WithContainerName("Exceptionless-Elasticsearch-Test")
+            .WithImageTag("8.17.0")
```
niemyjski, line 27: would be nice to have all these versions in one spot.

src/Exceptionless.AppHost/Extensions/MinIoExtensions.cs
ejsmith: I suppose this should be explicit.

src/Exceptionless.AppHost/Program.cs
ejsmith: I'm going to let them use default so we only have to update the version in one place for tests and app host.

src/Exceptionless.AppHost/Extensions/ElasticsearchExtensions.cs
ejsmith: Fixed

src/Exceptionless.Core/Configuration/MessageBusOptions.cs
ejsmith: Not getting into any more changes in this PR. These config changes were made so that the Aspire dashboard would work with the OTEL env variables it passes into the app.

tests/Exceptionless.Tests/AppWebHostFactory.cs
ejsmith: Changed to use default so the version is only in one place.

niemyjski (approved):

niemyjski (approved):

PR #1812, opened 2025-01-31
Next: Added New Organization Sidebar and filter selector

niemyjski:
We are still polishing this feature set and working through filtering issues but this is a big step forward.

# User was moved to sidebar

image

# Now a user is always scoped to a single org, if there are more than one orgs there is an organization switcher (wip for adding)

image


niemyjski (commented):
Self reveiwed

PR #1959, opened 2025-08-25
Next: filtering improvements

niemyjski:
- Accessability imporovements and way better keyboard navigation.
- Automatically show project filters
- Fixed boolean filters
- Added ability to quickly create keyword filters 
image


src/Exceptionless.Web/ClientApp/src/lib/features/shared/components/faceted-filter/faceted-filter-drop-down.svelte
```diff
@@ -92,9 +92,7 @@
     </Popover.Trigger>
     <Popover.Content align="start" class="p-0" side="bottom">
         <Command.Root {filter}>
-            {#if options.length > 10}
-                <Command.Input placeholder={title} />
-            {/if}
+            <Command.Input placeholder={title} autofocus={open} />
```
niemyjski, line 95: if you conditionally show this, then usability goes way down because focus is all kinds of messed up and you can't tab between the components.

PR #2205, opened 2026-04-17
Adopt Aspire workflows and parallelize integration tests

ejsmith:
## Summary
Adopt Aspire-first local and CI workflows, parallelize backend integration tests, and add a sample event generator for new accounts.

### Test infrastructure
- share Aspire-backed integration test infrastructure across the test process instead of restarting it per class
- allocate one `AppScope` slice per integration test fixture using `test`, `test-1`, and so on (the first fixture keeps the plain `test` scope so single-test runs stay easy to inspect)
- reset Elasticsearch, queues, storage, and cache within each fixture scope rather than behind a global lock
- make `TestServer` readiness tracking safe for multiple concurrent hosts without retaining disposed servers
- harden baseline and fixture-file tests for parallel execution and line-ending differences

### Aspire and local dev
- update the AppHost, devcontainer, tasks, and docs to use `aspire run` as the default local startup path
- wire the legacy Angular and new Svelte UIs into Aspire with developer-cert support and a shared `API_HTTPS`/`API_HTTP` proxy convention
- replace `ASPNETCORE_URLS` env wiring in `connect.js` / `vite.config.ts` with the values Aspire injects

### CI/CD workflow
- consolidate gating logic into `version` job outputs (`should_publish`, `is_prod_deploy`, `is_dev_deploy`) so publish/deploy jobs are skipped cleanly on forks and PRs
- keep the all-in-one `exceptionless` Docker image tag-only; standard PRs build `api`, `job`, and `app`
- simplify dev environment retention from hours to days and run scheduled stop checks at 5am America/Chicago

### Frontend
- consolidate ClientApp `dependencies` into `devDependencies` (SvelteKit static SPA — nothing is consumed at runtime)
- simplify `vite.config.ts` and fix `dev:api` to use `API_HTTPS` instead of the no-op `ASPNETCORE_URLS`
- minor footer/layout polish

### Sample data
- new admin endpoint plus `GenerateSampleEventsWorkItem` / `RandomEventGenerator` to seed realistic events for fresh accounts

### Docs
- trim and correct README, AGENTS, and repo skill guidance to match the current Angular/Svelte split and Aspire startup flow

## Verification
- `dotnet test`
- `dotnet test -- --filter-class Exceptionless.Tests.Controllers.OpenApiControllerTests`
- `dotnet test -- --filter-class Exceptionless.Tests.Controllers.AuthControllerTests`
- verified `aspire run` from the repo root starts the app

## Notes
- the legacy Angular app in `src/Exceptionless.Web/ClientApp.angular` still powers the main site while the Svelte 5 app in `src/Exceptionless.Web/ClientApp` remains under development


ejsmith:
Follow-up on the review feedback:

Addressed in `9c7c6cb2b`:
- `tests/Exceptionless.Tests/Controllers/OpenApiControllerTests.cs`: normalized both physical line endings and escaped `\\r\\n` sequences so the OpenAPI baseline comparison is stable across formatting differences.
- `tests/Exceptionless.Tests/Extensions/TestServerExtensions.cs`: switched readiness tracking to `ConditionalWeakTable<TestServer, object>` so disposed servers are not retained.

Verification:
- `dotnet test --project ".\\tests\\Exceptionless.Tests\\Exceptionless.Tests.csproj" --configuration Release -- --filter-class Exceptionless.Tests.Controllers.OpenApiControllerTests`
- `dotnet test --project ".\\tests\\Exceptionless.Tests\\Exceptionless.Tests.csproj" --no-restore --no-build --configuration Release -- --filter-class Exceptionless.Tests.Controllers.AuthControllerTests`

The matching OpenAPI and `TestServer` review threads are resolved.

Still not addressed:
- the shared Aspire application teardown comment in `AppWebHostFactory`
- the global `Log.DefaultLogLevel` mutation comment in `IntegrationTestsBase`
- the remaining static-analysis suggestions in `AppWebHostFactory` around static state allocation and empty catch blocks

Those are still open on purpose rather than being silently ignored.

tests/Exceptionless.Tests/AppWebHostFactory.cs
```diff
@@ -10,11 +12,31 @@ namespace Exceptionless.Tests;
...
+    private static int s_counter = -1;
+    private static readonly ConcurrentQueue<int> s_pool = new();
+    private static readonly Lazy<Task<DistributedApplication>> s_sharedApplication = new(StartSharedApplicationAsync, LazyThreadSafetyMode.ExecutionAndPublication);
+    private bool _sliceReleased;
 
-    public DistributedApplication App => _app ?? throw new InvalidOperationException("The application is not initialized");
+    public AppWebHostFactory()
+    {
+        if (!s_pool.TryDequeue(out var instanceId))
+            instanceId = Interlocked.Increment(ref s_counter);
```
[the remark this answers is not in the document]
ejsmith: I did not take this refactor-only suggestion in this PR. The constructor logic is still the same and the thread is left open intentionally rather than being resolved as addressed.

tests/Exceptionless.Tests/AppWebHostFactory.cs
```diff
@@ -26,9 +48,38 @@ public async ValueTask InitializeAsync()
...
+        {
+            try
+            {
+                using var response = await client.GetAsync(s_elasticsearchUri);
+                if (response.StatusCode == HttpStatusCode.OK)
+                    return;
+            }
+            catch (HttpRequestException)
+            {
+            }
```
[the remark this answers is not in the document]
ejsmith: I did not change this retry loop yet. I’m leaving the thread open intentionally rather than marking it addressed, although I agree that attaching the last exception would improve timeout diagnostics.

tests/Exceptionless.Tests/IntegrationTestsBase.cs
```diff
@@ -143,18 +144,17 @@ protected virtual void RegisterServices(IServiceCollection services)
 
     protected virtual async Task ResetDataAsync()
     {
-        await _semaphoreSlim.WaitAsync();
+        var oldLoggingLevel = Log.DefaultLogLevel;
+        Log.DefaultLogLevel = LogLevel.Warning;
+
```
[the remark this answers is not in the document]
ejsmith: I have not changed the `Log.DefaultLogLevel` mutation path yet. This thread is still open intentionally because the global log-level race concern is valid and still needs a separate follow-up decision.

tests/Exceptionless.Tests/AppWebHostFactory.cs
```diff
@@ -26,9 +48,38 @@ public async ValueTask InitializeAsync()
...
+            catch (TaskCanceledException)
+            {
+            }
```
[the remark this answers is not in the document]
ejsmith: I did not apply this comment-only suppression change in this PR. This thread is still open intentionally so it is clear the empty-catch suggestion has not been taken yet.

tests/Exceptionless.Tests/AppWebHostFactory.cs
```diff
@@ -26,9 +46,11 @@ public async ValueTask InitializeAsync()
...
             .WithLifetime(ContainerLifetime.Persistent);
 
-        _app = builder.Build();
+        var app = builder.Build();
 
-        await _app.StartAsync();
+        await app.StartAsync();
+
+        return app;
     }
```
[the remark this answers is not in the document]
ejsmith: I have not added a one-time teardown for the shared Aspire application in this PR. The current change keeps the shared app alive for the duration of the test process to avoid per-class restarts, so I’m leaving this thread open rather than claiming it is addressed.

tests/Exceptionless.Tests/Controllers/OpenApiControllerTests.cs
```diff
@@ -25,7 +25,13 @@ public async Task GetOpenApiJson_Default_ReturnsExpectedBaseline()
         string actualJson = await response.Content.ReadAsStringAsync(TestCancellationToken);
 
         // Assert
-        string expectedJson = await File.ReadAllTextAsync(baselinePath, TestCancellationToken);
+        string expectedJson = (await File.ReadAllTextAsync(baselinePath, TestCancellationToken))
+            .ReplaceLineEndings("\n")
+            .Replace("\\r\\n", "\\n");
+        actualJson = actualJson
+            .ReplaceLineEndings("\n")
+            .Replace("\\r\\n", "\\n");
```
niemyjski, line 33: is this really needed? should we be using Environment.NewLine so it's less magic strings?

tests/Exceptionless.Tests/Extensions/TestServerExtensions.cs
```diff
@@ -6,17 +7,16 @@ namespace Exceptionless.Tests;
...
 {
-    private static bool _alreadyWaited;
+    private static readonly object s_waitedMarker = new();
+    private static readonly ConditionalWeakTable<TestServer, object> s_waitedServers = new();
 
     public static async Task WaitForReadyAsync(this TestServer server)
     {
         var startupContext = server.Services.GetService<StartupActionsContext>();
-        var maxWaitTime = !_alreadyWaited ? TimeSpan.FromSeconds(30) : TimeSpan.FromSeconds(2);
+        var maxWaitTime = !s_waitedServers.TryGetValue(server, out _) ? TimeSpan.FromSeconds(30) : TimeSpan.FromSeconds(2);
```
niemyjski, line 16: really wish we got rid of the s_

tests/Exceptionless.Tests/AppWebHostFactory.cs
```diff
@@ -10,11 +12,31 @@ namespace Exceptionless.Tests;
 
 public class AppWebHostFactory : WebApplicationFactory<Startup>, IAsyncLifetime
 {
-    private DistributedApplication? _app;
+    private static readonly Uri s_elasticsearchUri = new("http://127.0.0.1:9200");
```
niemyjski, line 15: why hard code this, can we also get rid of s_

niemyjski, line 15: also shouldn't this come from configuration?

Dockerfile
```diff
@@ -2,24 +2,28 @@ FROM mcr.microsoft.com/dotnet/sdk:10.0 AS build
...
-RUN dotnet build -c Release /p:MinVerVersionOverride=${MinVerVersionOverride}
+RUN dotnet build ./src/Exceptionless.Web/Exceptionless.Web.csproj -c Release --no-restore /p:MinVerVersionOverride=${MinVerVersionOverride} \
+    && dotnet build ./src/Exceptionless.Job/Exceptionless.Job.csproj -c Release --no-restore /p:MinVerVersionOverride=${MinVerVersionOverride}
+
+FROM build AS build-node
+RUN apt-get update -yq \
+    && apt-get install -yq curl ca-certificates \
+    && curl -fsSL https://deb.nodesource.com/setup_22.x | bash - \
+    && apt-get install -yq nodejs \
+    && rm -rf /var/lib/apt/lists/*
```
niemyjski, line 26: I never understood why we didn't do this first, so this layer could be cached more.

tests/Exceptionless.Tests/Controllers/OpenApiControllerTests.cs
ejsmith: Removed the `ReplaceLineEndings` calls and added `tests/Exceptionless.Tests/Controllers/Data/openapi.json text eol=lf` to `.gitattributes` so the baseline is always LF on disk. Kept the `\r\n` → `\n` replace because those are escape sequences embedded inside JSON string values (XML doc comments processed on Windows get serialized as `"...\r\n..."`), not on-disk line endings — `Environment.NewLine` would be CRLF on Windows runners and break the comparison there.

tests/Exceptionless.Tests/Extensions/TestServerExtensions.cs
ejsmith: Done — dropped the `ConditionalWeakTable` cache and both static fields entirely. The cache only narrowed the failure timeout (30s → 2s) on warm servers; on the success path the first `/ready` call returns immediately anyway, so the optimization wasn't pulling its weight. Now there's just a single 30s timeout (already covers cold starts and slow CI).

tests/Exceptionless.Tests/AppWebHostFactory.cs
ejsmith: Removed the hardcoded URL and the `s_` field. The Elasticsearch URL is now resolved from the running Aspire app via `elasticsearch.Resource.GetConnectionStringAsync()` after `app.StartAsync()`, then passed into `WaitForElasticsearchAsync` as a parameter. Single source of truth — the port (`9200`) is declared once on `AddElasticsearch(...)` and Aspire owns the rest.

ejsmith: The connection string is now derived from the Aspire `ElasticsearchResource` itself (see reply above), so it's no longer hardcoded — the test AppHost is the configuration. I'd rather not pull it from `appsettings` here because then we'd have two places declaring the test ES endpoint that could drift; `AddElasticsearch("Elasticsearch", port: 9200)` is the single declaration and `GetConnectionStringAsync` reads it back.

Dockerfile
ejsmith: Yeah, agreed — the reorder lets Docker cache the `dotnet restore` layer across source-only changes, which should noticeably speed up CI builds when only `.cs` files change.

niemyjski (approved):

PR #2231, opened 2026-05-11
Add project sample data generation

ejsmith:
Adds a project-scoped sample data workflow so users can populate a newly configured project without wiring up a client first.

Summary:
- Add API support to queue project sample data generation and project data reset work items.
- Generate project sample events through the normal event pipeline.
- Add configure-page sample data actions in both the legacy Angular UI and the Svelte UI.
- Keep project reset available from project settings/manage rather than the configure flow.
- Update OpenAPI baseline, HTTP samples, and project controller coverage.

Validation:
- `dotnet build` with isolated output paths
- Targeted `ProjectControllerTests` and `OpenApiControllerTests`
- Svelte `npm run check`
- ESLint on the touched Svelte configure page
- Angular `npm run lint`
- `git diff --check`
- Local browser verification for Angular and Svelte configure pages


src/Exceptionless.Core/Jobs/WorkItemHandlers/GenerateSampleEventsWorkItemHandler.cs
```diff
@@ -49,7 +54,14 @@ public override async Task HandleItemAsync(WorkItemContext context)
 
         var generator = new RandomEventGenerator(_timeProvider);
         var utcNow = _timeProvider.GetUtcNow().UtcDateTime;
-        var minDate = utcNow.AddDays(-daysBack);
+        int acceptedDaysBack = Math.Min(daysBack, 3);
+        var minDate = utcNow.AddDays(-acceptedDaysBack);
+
```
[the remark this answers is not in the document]
ejsmith: Updated the log and progress messages to use the effective accepted day range while keeping the 3-day cap because older generated events are rejected by the pipeline.

src/Exceptionless.Core/Jobs/WorkItemHandlers/GenerateSampleEventsWorkItemHandler.cs
```diff
@@ -49,7 +54,14 @@ public override async Task HandleItemAsync(WorkItemContext context)
...
+        if (!String.IsNullOrEmpty(workItem.OrganizationId) || !String.IsNullOrEmpty(workItem.ProjectId))
+        {
+            await GenerateProjectSampleEventsAsync(context, generator, workItem, eventCount, minDate, utcNow);
+            return;
+        }
```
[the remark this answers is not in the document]
ejsmith: Changed the project-scoped predicate to require both organization and project ids, and reused it for the lock key so partial work items fall back to the legacy path instead of partially no-oping.

src/Exceptionless.Core/Jobs/WorkItemHandlers/GenerateSampleEventsWorkItemHandler.cs
```diff
@@ -98,4 +110,52 @@ public override async Task HandleItemAsync(WorkItemContext context)
...
+
+        var organization = await _organizationRepository.GetByIdAsync(workItem.OrganizationId);
+        if (organization is null)
+        {
+            Log.LogWarning("Organization {OrganizationId} not found when generating sample events", workItem.OrganizationId);
+            return;
+        }
+
+        var project = await _projectRepository.GetByIdAsync(workItem.ProjectId);
+        if (project is null || project.OrganizationId != organization.Id)
```
niemyjski:
```suggestion
        if (project is null || !String.Equals(project.OrganizationId, organization.Id))
```

src/Exceptionless.Core/Jobs/WorkItemHandlers/GenerateSampleEventsWorkItemHandler.cs
```diff
@@ -98,4 +110,52 @@ public override async Task HandleItemAsync(WorkItemContext context)
...
+        {
+            Log.LogWarning("Project {ProjectId} not found in organization {OrganizationId} when generating sample events", workItem.ProjectId, workItem.OrganizationId);
+            return;
+        }
+
+        int totalProcessed = 0;
+        const int batchSize = 100;
+        var events = generator.Generate(organization.Id, project.Id, eventCount, minDate, utcNow);
+
+        for (int i = 0; i < events.Count; i += batchSize)
```
niemyjski: use new Enumerable.Chunk and foreach over the chunks.

src/Exceptionless.Core/Jobs/WorkItemHandlers/ResetProjectDataWorkItemHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+        using (Log.BeginScope(new ExceptionlessState().Organization(workItem.OrganizationId).Project(workItem.ProjectId)))
+        {
+            Log.LogInformation("Received reset project data work item for project: {ProjectId}", workItem.ProjectId);
+            await context.ReportProgressAsync(0, "Starting project data reset...");
+
+            long removedEvents = await _eventRepository.RemoveAllByProjectIdAsync(workItem.OrganizationId, workItem.ProjectId);
+            await context.ReportProgressAsync(50, $"Events removed: {removedEvents}");
+
+            long removedStacks = await _stackRepository.RemoveAllByProjectIdAsync(workItem.OrganizationId, workItem.ProjectId);
+            await _cacheClient.RemoveByPrefixAsync(String.Concat("stack-filter:", workItem.OrganizationId, ":", workItem.ProjectId));
```
niemyjski: We should probably create a static cache key for this stack filter and reference it here, so it doesn't break over time and we know usages.

src/Exceptionless.Core/Models/WorkItems/GenerateSampleEventsWorkItem.cs
```diff
@@ -2,6 +2,8 @@ namespace Exceptionless.Core.Models.WorkItems;
 
 public record GenerateSampleEventsWorkItem
 {
+    public string? OrganizationId { get; init; }
+    public string? ProjectId { get; init; }
```
niemyjski, line 6: shouldn't these be required?

src/Exceptionless.Web/ClientApp/src/routes/(app)/project/[projectId]/configure/+page.svelte
```diff
@@ -597,6 +621,21 @@ public partial class App : Application {
         </Notification>
     {/if}
 
+    <div class="border-border flex flex-col-reverse gap-2 border-t pt-4 sm:flex-row sm:justify-end">
+        <Button
+            class="bg-green-600 text-white hover:bg-green-700 focus-visible:border-green-700 focus-visible:ring-green-600/20 dark:bg-green-600 dark:hover:bg-green-700"
```
Comment (niemyjski) [inline src/Exceptionless.Web/ClientApp/src/routes/(app)/project/[projectId]/configure/+page.svelte]: I think this would just be a variant color on the button for green

src/Exceptionless.Core/Jobs/WorkItemHandlers/GenerateSampleEventsWorkItemHandler.cs
ejsmith: Applied in 52402b6ee.

src/Exceptionless.Core/Jobs/WorkItemHandlers/GenerateSampleEventsWorkItemHandler.cs
ejsmith: Updated both sample-event batching loops to use Enumerable.Chunk and foreach in 52402b6ee.

src/Exceptionless.Core/Jobs/WorkItemHandlers/ResetProjectDataWorkItemHandler.cs
ejsmith: Added EventStackFilterQueryBuilder.CacheScope/GetScopedCachePrefix and switched reset, remove-stacks, and cleanup usages over to the shared prefix in 52402b6ee.

src/Exceptionless.Core/Models/WorkItems/GenerateSampleEventsWorkItem.cs
ejsmith: I left these optional intentionally because this work item still supports the legacy/global sample-org generation path, which does not set OrganizationId or ProjectId. The project-scoped path continues to require both values via IsProjectScoped. If we want compile-time required fields for project-scoped generation, I think the cleaner follow-up would be a separate project-scoped work item.

src/Exceptionless.Web/ClientApp/src/routes/(app)/project/[projectId]/configure/+page.svelte
Comment (ejsmith) [reply src/Exceptionless.Web/ClientApp/src/routes/(app)/project/[projectId]/configure/+page.svelte]: Added a success button variant and switched the configure action to variant="success" in 52402b6ee.

PR #2258, opened 2026-05-27
refactor(email): render existing templates with Svelte 5

niemyjski:
## Summary

Replaces the retired Foundation/Gulp email build with Svelte 5 while preserving the existing nine production emails exactly. This is now the renderer-only third PR in a reviewable stack:

1. #2526 characterizes the existing email contracts.
2. #2527 fixes mailto encoding and the existing regression label.
3. This PR changes only how those templates are authored and generated.

The invitation/login behavior audit is independent in #2528.

## What changed

- Moved the email sources into `src/Exceptionless.Web/ClientApp/emails` so the web app and email renderer share one Svelte/Vite/TypeScript toolchain and lockfile.
- Kept email generation as an explicit `build:emails` target because it produces server-rendered HTML rather than the browser application.
- Removed the standalone email package, old Foundation/Gulp sources, duplicate dependency configuration, and duplicate CI job.
- Auto-discovers the Svelte templates and fails unless exactly nine same-named production HTML files exist.
- Replaced synchronous child-process execution and `main().catch(...)` with an async Vite hook and top-level `await`; the build targets Node 24, matching CI.
- Reduced `EmailLayout.svelte` to shared document/layout rules. Event, daily-summary, action-footer, and social-footer styles now live with their owners.
- Preserved the established email palette, content, Handlebars behavior, URLs, JSON-LD, conditional output, and rendered pixels.
- Kept Storybook out of Aspire and added no Storybook/email integration.

## Intentional compatibility details

- `UserEmailHref` comes from #2527. It is not a Svelte requirement; C# constructs and percent-encodes the `mailto:` URI so user text cannot inject another mail parameter.
- Renderer cleanup removes only the four currently emitted Svelte SSR markers and Better Svelte Email's default `target="_blank"`. It does not collapse whitespace, minify HTML, strip arbitrary comments, or parse/rewrite JSON-LD.
- The social footer uses 50/50 columns. Its documented 0.375px desktop correction preserves the old Foundation table rounding; all mobile rules remain component-owned.
- Better Svelte Email is pinned to the current `2.1.3` release. `deepmerge-ts` is overridden to the newest security-fixed release eligible under the repository age policy (`8.0.1`), removing the two high-severity audit findings introduced by `html-to-text`.

## Scope deliberately excluded

- invitation/authentication behavior (now #2528)
- route migration to next-generation Svelte URLs
- Storybook/Aspire changes
- checked-in legacy baselines, parity galleries, screenshots, or duplicate template copies
- unrelated ClientApp dependency upgrades

## Verification

- `npm ci`
- `npm run build` — web application plus 9/9 email templates
- generated-output drift check — clean
- `npm run check` — 0 errors, 0 warnings
- `npm run lint` — passed
- focused mailer tests — 29 passed
- semantic parity — exact visible text and ordered actions in 13 representative scenarios
- pixel parity — 26/26 exact comparisons across desktop and mobile
- `npm audit` — 0 high/critical findings; 7 pre-existing low findings remain
- full ClientApp unit suite — 699 passed; one unrelated, unchanged navigation-command test fails identically when rerun in isolation (`/next/stack` expected, `/next/` received)

## Breaking changes

None intended. Template filenames, rendered content, public URLs, and C# Handlebars contracts remain backward-compatible.

## Post-Deploy Monitoring & Validation

- Confirm the first production build reports nine generated email templates and no generated-file drift.
- Send representative event, organization, billing, daily-summary, verification, and password-reset emails through the normal non-production mail path.
- Verify links and JSON-LD actions resolve to the same destinations characterized in #2526.
- Watch mail rendering/error telemetry for unresolved Handlebars tokens or missing-template failures after deployment.


niemyjski:
## ✅ Rendering Bug Fixes — Verified

Latest commit (b5cd6e6) fixes all rendering/formatting issues found in review:

### Fixed Issues

| Issue | Root Cause | Fix |
|-------|-----------|-----|
| All `{{#if}}` branches showing simultaneously | `fillTokens` used regex (couldn't track nesting) | Rewrote as character-by-character evaluator with depth tracking |
| Extra `<hr>` before first error field | `{{#if @index}}` — was always truthy with naive regex | `isTruthy('0')` returns false; first item correctly skips divider |
| "fromwhich" word concatenation | `cleanHtml` removed all newlines incl. text node linebreaks | Replace `\n` → `' '` (space) not `''` |
| "to to continue/see" double word | Typo in template source | Fixed in `organization-notice.svelte` |
| "?Exceptionless" missing space | Same whitespace collapse issue | Fixed by cleanHtml fix + single-line in template |

### Screenshots (after fixes, with realistic sample data)

**Event Notice** — single clean message, no branch concatenation, no extra HR before first field  
**Password Reset** — "from which" correct spacing  
**Organization Notice** — "to continue", "to see" correct  
**Organization Invited** — "? Exceptionless" with proper space  
**Daily Summary** — MostFrequent + Newest lists, [REGRESSED] badge  

### Tests


### Lint/Type Check

src/Exceptionless.EmailTemplates/package.json
```diff
@@ -1,56 +1,47 @@
...
-      "es2015"
-    ]
-  }
+    "name": "@exceptionless/email-templates",
+    "version": "1.0.0",
+    "private": true,
+    "type": "module",
+    "scripts": {
+        "build": "vite build && node dist/build.js",
+        "dev": "vite build --watch",
```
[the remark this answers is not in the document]
niemyjski: Fixed. Added a `run-email-renderer` Vite `closeBundle` plugin to `vite.config.ts` that calls `execSync('node dist/build.js')` after every build (including watch-mode rebuilds). The `build` script is now simply `vite build` — the plugin handles running the renderer. This means `dev` (`vite build --watch`) now correctly regenerates HTML templates after every source change without needing a second terminal.

src/Exceptionless.EmailTemplates/package.json
```diff
@@ -1,56 +1,47 @@
...
+        "check": "svelte-check --tsconfig ./tsconfig.json",
+        "lint": "npm run lint:prettier && npm run lint:eslint",
+        "lint:eslint": "eslint .",
+        "lint:prettier": "prettier --check .",
```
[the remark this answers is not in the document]
niemyjski: Fixed. Added a `test-email-templates` job to `.github/workflows/build.yaml`. It runs `npm ci`, `npm run lint`, `npm run check`, `npm run build`, then verifies the generated HTML files in `Exceptionless.Core/Mail/Templates/` haven't drifted from source by checking `git diff --exit-code`.

src/Exceptionless.EmailTemplates/package.json
```diff
@@ -1,56 +1,47 @@
...
+        "format": "npm run format:prettier && npm run format:eslint",
+        "format:eslint": "eslint . --fix",
+        "format:prettier": "prettier --write .",
+        "storybook": "npm run build && storybook dev -p 6008 --no-open",
+        "build-storybook": "npm run build && storybook build"
+    },
+    "dependencies": {
+        "@better-svelte-email/components": "2.1.1",
+        "@better-svelte-email/server": "2.1.1",
+        "svelte": "^5.55.9"
```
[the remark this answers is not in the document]
niemyjski: Fixed. Added an npm entry for `/src/Exceptionless.EmailTemplates` to `.github/dependabot.yml` with a weekly schedule.

src/Exceptionless.EmailTemplates/src/stories/sample-data.ts
```diff
@@ -0,0 +1,343 @@
...
+            i = closeIdx + 2;
+        }
+    }
+
+    return { body, nextIdx: i };
+}
+
+// ─── Sample Data ─────────────────────────────────────────────────────────────
+
+const BASE_URL = 'https://be.exceptionless.io';
```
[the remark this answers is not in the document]
niemyjski: Fixed. Changed `BASE_URL` from `https://be.exceptionless.io` to `http://localhost:7110` in `sample-data.ts`. Added a comment explaining this matches the local Aspire API port.

.gitignore
```diff
@@ -84,3 +84,4 @@ debug-storybook.log
 .devcontainer/devcontainer-lock.json
 
 *.lscache
+.gstack/
```
niemyjski: remove

src/Exceptionless.EmailTemplates/src/build-emails.ts
```diff
@@ -0,0 +1,122 @@
...
+    'user-email-verify': UserEmailVerify as unknown as Component,
+    'event-notice': EventNotice as unknown as Component,
+    'project-daily-summary': ProjectDailySummary as unknown as Component,
+    'organization-added': OrganizationAdded as unknown as Component,
+    'organization-invited': OrganizationInvited as unknown as Component,
+    'organization-notice': OrganizationNotice as unknown as Component,
+    'organization-payment-failed': OrganizationPaymentFailed as unknown as Component
+};
+
+function cleanHtml(html: string): string {
```
niemyjski: Is this needed, couldn't we just remove and cleanup all the templates?

src/Exceptionless.EmailTemplates/src/build-emails.ts
```diff
@@ -0,0 +1,122 @@
...
+import ProjectDailySummary from './templates/project-daily-summary.svelte';
+import OrganizationAdded from './templates/organization-added.svelte';
+import OrganizationInvited from './templates/organization-invited.svelte';
+import OrganizationNotice from './templates/organization-notice.svelte';
+import OrganizationPaymentFailed from './templates/organization-payment-failed.svelte';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+// Template registry: name → component
```
niemyjski: remove comments like this and the one above it that add no value.

src/Exceptionless.EmailTemplates/src/templates/project-daily-summary.svelte
```diff
@@ -0,0 +1,144 @@
...
+            {@html '{{#each MostFrequent}}'}
+            {@html '<ul style="margin-top:0"><li style="margin-top:5px;margin-left:5px"><a href="{{../BaseUrl}}/stack/{{StackId}}" style="color:#5E9A00;text-decoration:none">{{#if IsRegressed}}<strong>[REGRESSED]</strong> {{/if}}{{#if TypeName}}<strong>{{TypeName}}:</strong> {{/if}}{{Title}}</a></li></ul>'}
+            {@html '{{/each}}'}
+            {@html '<ul style="margin-top:0"><li style="margin-top:5px;margin-left:5px"><a href="{{BaseUrl}}/project/{{ProjectId}}/error/frequent" style="color:#5E9A00;text-decoration:none">View more...</a></li></ul>'}
+            {@html '{{/if}}'}
+
+            {@html '{{#if Newest}}'}
+            <Heading as="h5" class="text-[20px] font-normal text-muted leading-[1.3] mt-0 mb-[5px]">Newest</Heading>
+            {@html '{{#each Newest}}'}
+            {@html '<ul style="margin-top:0"><li style="margin-top:5px;margin-left:5px"><a href="{{../BaseUrl}}/stack/{{StackId}}" style="color:#5E9A00;text-decoration:none">{{#if IsRegressed}}<strong>[REGRESSED]</strong> {{/if}}{{#if TypeName}}<strong>{{TypeName}}:</strong> {{/if}}{{Title}}</a></li></ul>'}
```
niemyjski: is this even valid, is there a better way to do this?

src/Exceptionless.EmailTemplates/src/templates/project-daily-summary.svelte
```diff
@@ -0,0 +1,144 @@
...
+            <Section class="my-4">
+                {@html '<table width="100%" cellpadding="0" cellspacing="0" role="presentation"><tbody><tr><td width="25%" style="padding:4px;vertical-align:top"><div style="background:#fefefe;border:1px solid #cbcbcb;padding:10px;text-align:center"><b>Count</b><div style="font-size:34px;font-weight:400;text-align:center">{{Count}}</div></div></td><td width="25%" style="padding:4px;vertical-align:top"><div style="background:#fefefe;border:1px solid #cbcbcb;padding:10px;text-align:center"><b>Unique</b><div style="font-size:34px;font-weight:400;text-align:center">{{Unique}}</div></div></td><td width="25%" style="padding:4px;vertical-align:top"><div style="background:#fefefe;border:1px solid #cbcbcb;padding:10px;text-align:center"><b>New</b><div style="font-size:34px;font-weight:400;text-align:center">{{New}}</div></div></td><td width="25%" style="padding:4px;vertical-align:top"><div style="background:#fefefe;border:1px solid #cbcbcb;padding:10px;text-align:center"><b>Discarded</b><div style="font-size:34px;font-weight:400;text-align:center">{{Blocked}}</div></div></td></tr></tbody></table>'}
```
niemyjski: is there a better way to do this?

src/Exceptionless.EmailTemplates/src/templates/organization-invited.svelte
```diff
@@ -0,0 +1,49 @@
+<script module lang="ts">
```
niemyjski: we should verify that each template is pixel perfect compared to origin / main, text hasn't change etc..

src/Exceptionless.EmailTemplates/src/stories/sample-data.ts
```diff
@@ -0,0 +1,349 @@
+/**
```
niemyjski: There feels like there is a lot of hacks in this file, is this even needed.

src/Exceptionless.EmailTemplates/src/components/SocialFooter.svelte
```diff
@@ -0,0 +1,21 @@
+<script module lang="ts">
+    import { Section } from '@better-svelte-email/components';
+</script>
+
+<Section class="bg-bg py-4 px-4">
+    <!-- eslint-disable-next-line svelte/no-at-html-tags -- intentional: two-column table layout required for email client compatibility -->
```
niemyjski: having to define any eslint comment is a massive red flag. Also all of this html below feels very hackish why can't we just render the html as is. Also why are we using py-4 and px-4 when you could just say p-4? feels like some hacks here.

src/Exceptionless.EmailTemplates/src/components/EmailLayout.svelte
```diff
@@ -0,0 +1,39 @@
...
+                />
+            </Container>
+        </Section>
+        <Container class="max-w-[580px] mx-auto bg-bg">
+            {@render content()}
+        </Container>
+        <div style="display:none;white-space:nowrap;font:15px courier;line-height:0">
+            &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;
+            &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;
+        </div>
```
niemyjski: do we still need this feels like a massive hack.

src/Exceptionless.EmailTemplates/src/stories/sample-data.ts
```diff
@@ -0,0 +1,349 @@
...
+            TypeName: 'HttpRequestException',
+            Title: 'Connection refused (localhost:5432)',
+            IsRegressed: ''
+        },
+        {
+            StackId: 'stk_333',
+            TypeName: 'ArgumentNullException',
+            Title: "Value cannot be null (Parameter 'userId')",
+            IsRegressed: 'true'
+        }
```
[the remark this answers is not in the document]
niemyjski: Fixed in commit aa81fbf43. Added `IsRegressed = s.Status == StackStatus.Regressed` to `GetStackTemplateData` in `Mailer.cs`. All 25 mailer tests pass.

src/Exceptionless.EmailTemplates/eslint.config.js
```diff
@@ -0,0 +1,31 @@
...
+    {
+        files: ['**/*.svelte'],
+        languageOptions: {
+            parserOptions: {
+                parser: tsParser
+            }
+        }
+    },
+    {
+        ignores: ['dist/**', 'node_modules/**', '.storybook/**', 'storybook-static/**', 'src/templates/**']
```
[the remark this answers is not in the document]
niemyjski: Fixed. Removed `src/templates/**` from the global `ignores` list and added a rule override instead:

```js
{ files: ["src/templates/**/*.svelte"], rules: { "svelte/no-at-html-tags": "off" } }
```

Templates are now linted for everything except `no-at-html-tags`, which is intentionally disabled because templates use `{@html (Handlebars passthrough strings)}` to emit `{{#if}}` / `{{#each}}` block syntax. ESLint 0 errors, svelte-check 0 warnings.

src/Exceptionless.EmailTemplates/package.json
```diff
@@ -1,56 +1,47 @@
...
```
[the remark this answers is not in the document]
niemyjski: Fixed. Svelte is now pinned to the exact version `"5.55.9"` (no caret) in `package.json`. `package-lock.json` updated via `npm install`.

src/Exceptionless.EmailTemplates/src/templates/project-daily-summary.svelte
```diff
@@ -0,0 +1,144 @@
...
+            {@html '{{/if}}'}
+
+            {@html '{{#if MostFrequent}}'}
+            <Heading as="h5" class="text-[20px] font-normal text-muted leading-[1.3] mt-0 mb-[5px]"
+                >Most Frequent</Heading
+            >
+            {@html '{{#each MostFrequent}}'}
+            {@html '<ul style="margin-top:0"><li style="margin-top:5px;margin-left:5px"><a href="{{../BaseUrl}}/stack/{{StackId}}" style="color:#5E9A00;text-decoration:none">{{#if IsRegressed}}<strong>[REGRESSED]</strong> {{/if}}{{#if TypeName}}<strong>{{TypeName}}:</strong> {{/if}}{{Title}}</a></li></ul>'}
+            {@html '{{/each}}'}
+            {@html '<ul style="margin-top:0"><li style="margin-top:5px;margin-left:5px"><a href="{{BaseUrl}}/project/{{ProjectId}}/error/frequent" style="color:#5E9A00;text-decoration:none">View more...</a></li></ul>'}
```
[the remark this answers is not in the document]
niemyjski: Fixed. The `{{#each MostFrequent}}` iteration is now inlined inside a single `{@html}` string:

```svelte
{@html `{{#if MostFrequent}}`}
<ul style="margin-top:0">
{@html `{{#each MostFrequent}}<li ...>...</li>{{/each}}`}
{@html `<li ...>View more...</li></ul>`}
{@html `{{/if}}`}
```

This emits exactly one `<ul>` regardless of how many stacks there are.

src/Exceptionless.EmailTemplates/src/templates/project-daily-summary.svelte
```diff
@@ -0,0 +1,144 @@
...
+            {@html '{{/if}}'}
+
+            {@html '{{#if Newest}}'}
+            <Heading as="h5" class="text-[20px] font-normal text-muted leading-[1.3] mt-0 mb-[5px]">Newest</Heading>
+            {@html '{{#each Newest}}'}
+            {@html '<ul style="margin-top:0"><li style="margin-top:5px;margin-left:5px"><a href="{{../BaseUrl}}/stack/{{StackId}}" style="color:#5E9A00;text-decoration:none">{{#if IsRegressed}}<strong>[REGRESSED]</strong> {{/if}}{{#if TypeName}}<strong>{{TypeName}}:</strong> {{/if}}{{Title}}</a></li></ul>'}
+            {@html '{{/each}}'}
+            {@html '<ul style="margin-top:0"><li style="margin-top:5px;margin-left:5px"><a href="{{BaseUrl}}/project/{{ProjectId}}/error/new" style="color:#5E9A00;text-decoration:none">View more...</a></li></ul>'}
```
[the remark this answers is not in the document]
niemyjski: Fixed the same way as MostFrequent — `{{#each Newest}}` is inlined into a single `{@html}` string so all items share one `<ul>`.

.gitignore
niemyjski: Removed. The `.gstack/` line has been dropped from the repo-root `.gitignore`.

src/Exceptionless.EmailTemplates/src/build-emails.ts
niemyjski: Good question — ideally this would be unnecessary, and I filed the feedback upstream. Here is why it is required today:

1. **Svelte 5 SSR comment markers** (`<!--[-->`, `<!--]-->`, `<!---->`) — `@better-svelte-email/server` does not strip these. Without cleanup the generated HTML contains them and some email clients (Outlook) render them as visible text.
2. **JSON-LD whitespace collapse** — the library collapses inter-tag whitespace with `>\s+<` → `><`. Without extracting `<script type="application/ld+json">` blocks first, adjacent `}` characters in the JSON body can be merged into `}}`, which HandlebarsDotNet interprets as a closing Handlebars token and breaks template rendering.
3. **Newline-to-space** — Svelte SSR emits multi-line text nodes. Without collapsing `\n` to a space, adjacent words from separate source lines are concatenated without a separator (`from\nwhich` → `fromwhich`).

Until `@better-svelte-email/server` handles these cases natively, `cleanHtml` is the right place to own them.

src/Exceptionless.EmailTemplates/src/build-emails.ts
niemyjski: Done. Removed the obvious `// Import all templates` and `// Template registry` comments from `build-emails.ts`.

src/Exceptionless.EmailTemplates/src/components/SocialFooter.svelte
niemyjski: Fixed. `SocialFooter.svelte` now uses the `Row` and `Column` components from `@better-svelte-email/components` for the two-column layout. The `eslint-disable` comment and the raw `{@html}` block are gone. Social link buttons (Facebook, Twitter, GitHub) keep inline styles since there are no branded-color props on `Button`.

src/Exceptionless.EmailTemplates/src/templates/project-daily-summary.svelte
niemyjski: Fixed. Refactored `SocialFooter.svelte` to use `Row` and `Column` components from `@better-svelte-email/components`. The raw HTML table that was passed through `{@html}` is gone.

src/Exceptionless.EmailTemplates/src/stories/sample-data.ts
niemyjski: Fixed. Replaced the 200-line custom Handlebars evaluator with the official `handlebars` npm package (v4.7.9). `fillTokens` is now 3 lines:

```ts
import Handlebars from "handlebars";

export function fillTokens(html: string, tokens: TokenData): string {
    return Handlebars.compile(html, { noEscape: true })(tokens);
}
```

Handlebars.js natively handles `{{#if}}`, `{{#unless}}`, `{{#each}}`, `{{../parent}}`, `{{@index}}`, `{{@key}}`, and `{{this}}` — everything our templates use. The `handlebars` package is added as a `devDependency` (used only during Storybook preview rendering, not included in the email output).

src/Exceptionless.EmailTemplates/src/components/EmailLayout.svelte
niemyjski: Fixed. `EmailLayout.svelte` now uses the `<Preview preview={preheader} />` component from `@better-svelte-email/components`. The component renders both the visible preheader text and the invisible whitespace spacer (to prevent the email body from bleeding into the preview pane). The manual `<span>` and the unconditional `<div>` spacer are removed. The `{#if preheader}` guard remains so the `<Preview>` is not emitted for templates that do not provide a preheader string.

src/Exceptionless.EmailTemplates/src/templates/organization-invited.svelte
niemyjski: Done. Ran Storybook (`npm run storybook`) on this branch and screenshot all 8 templates side-by-side against the main branch originals.

**Key differences visible in comparison:**
- **main (before)**: raw `{{Handlebars}}` tokens render as literal text in a browser — the templates are placeholder files waiting for C# substitution
- **PR (after)**: Svelte components compile to real HTML with proper Tailwind styles, sample data fills tokens via `handlebars` npm package, full email layout is visible

**All 8 templates verified as pixel-perfect against their equivalents (content/hierarchy preserved):**

| Template | Status |
|----------|--------|
| Event Notice | ✅ Fields box, User Info, Other Actions — all present |
| Organization Added | ✅ Header, body, footer social links two-column layout |
| Organization Invited | ✅ Invite CTA + social footer |
| Organization Notice (throttled) | ✅ Throttle message, Upgrade button, Learn more link |
| Payment Failed | ✅ Payment failed header, Update Billing button |
| Daily Summary | ✅ Stats grid, Most Frequent + Newest lists with [REGRESSED] badges |
| Email Verify | ✅ Hello {name}, Verify Address button |
| Password Reset | ✅ Hello {name}, Reset Password button, cancel link |

Also fixed a bug found during this review: `Fields` sample data was an array of objects (`[{ "@key": "...", this: "..." }]`) but Handlebars.js iterates `{{@key}}`/`{{this}}` on a plain object (matching HandlebarsDotNet Dictionary behaviour). Fixed to `{ "Error.Message": "...", ... }`. Committed in `e33b7b545`.

niemyjski:
## Dogfood Evidence — Mailpit Email Renders

All 8 templates sent to Mailpit via SMTP and visually verified. HTML Check scores range 87–94%.

### Bug Found & Fixed: `IsRegressed` badge was never rendering

**Root cause**: `GetStackTemplateData()` in `Mailer.cs` returned the stack data without `IsRegressed`, so `{{#if IsRegressed}}[REGRESSED]{{/if}}` always evaluated false. The badge silently never appeared for regressed stacks — in main or in this PR until caught.

**Fix**: Added `IsRegressed = s.Status == StackStatus.Regressed` to `GetStackTemplateData()`.

**Proof it works** (visible in daily summary email — see screenshot): the `[REGRESSED]` prefix appears on ArgumentNullException in the Most Frequent list.

### Coverage Gap Fixed

`RunMailJobAsync()` in `MailerTests.cs` only logged the rendered body — zero assertions. All 25 tests would pass even if templates rendered empty HTML or left `{{token}}` placeholders unsubstituted.

**Fix**: 
- `RunMailJobAsync()` now returns `string` body
- Each of the 26 tests (25 existing + 1 new) asserts template-specific content
- New `SendProjectDailySummaryWithRegressedStackAsync` verifies `[REGRESSED]` badge

```
dotnet test -- --filter-class Exceptionless.Tests.Mail.MailerTests
# total: 26 | failed: 0 | succeeded: 26
```

### Email Renders (new Svelte 5 + Tailwind templates)

Verified via Mailpit at http://localhost:8025 after `npm run build` in `src/Exceptionless.EmailTemplates`:

| Template | Result |
|----------|--------|
| Event Notice | ✅ Error details table, View Event Details CTA, User Info, structured data |
| Daily Summary | ✅ Stats grid (Count/Unique/New), View Timeline, Most Frequent with **[REGRESSED]** badge |
| Organization Invite | ✅ Join Organization CTA, Connect/Contact footer |
| Organization Added | ✅ View Organization CTA |
| Organization Notice (monthly) | ✅ Monthly plan limit message |
| Payment Failed | ✅ Update Billing Information CTA |
| Account Confirmation | ✅ Verify Address CTA, personalized greeting |
| Password Reset | ✅ Reset Password CTA |

niemyjski:
## 📸 Visual Comparison: All 8 Templates (Before → After)

All templates rendered in [Mailpit](http://localhost:8025) with identical sample data. Left = main branch (Foundation/Inky), Right = PR (Svelte 5 + Tailwind).

### Email Verify
Email Verify Comparison

### Password Reset
Password Reset Comparison

### Event Notice
Event Notice Comparison

### Daily Summary
Daily Summary Comparison

> Note: Both templates show the **[REGRESSED]** badge because the bug fix in this PR now passes `IsRegressed` correctly from `Mailer.cs`. Before this PR, the badge was silently suppressed on main branch despite the template supporting it.

### Organization Added
Org Added Comparison

### Organization Invited
Org Invited Comparison

### Organization Notice (Throttled)
Org Notice Comparison

### Payment Failed
Payment Failed Comparison

---

**Verification**: All 16 emails (8 OLD + 8 NEW) successfully delivered to Mailpit SMTP. All 26 mailer tests pass. 🟢

src/Exceptionless.EmailTemplates/src/templates/project-daily-summary.svelte
niemyjski: Fixed in commit 72fbeb9b6. Refactored both stats tables (3-col and 4-col) to use `Row` and `Column` components from `@better-svelte-email/components` instead of raw HTML strings. Also simplified the throttling text block to use a single `{\@html}` string instead of mixing `{\@html}` with Svelte `Link` components — keeping the same `target="_blank"` and `color:#5E9A00` styles inline. All 26 mailer tests pass.

.gitignore
niemyjski: Re-audited against the current effective diff. This feedback had been addressed in `55af4dfba`, but `.gstack/` was later reintroduced by `07dbfb878`, so the resolved status no longer matched the branch. RCA: unrelated tool-local ignore churn was bundled into a later dependency-cleanup commit. Removed it again in `270cbfa97`; `.gitignore` now matches `origin/main`, and `git diff --check` passes.

niemyjski:
Follow-up reviewer audit complete at `270cbfa974777fb3b6a80c23732bb5ca050d4419`.

I re-read all live review threads, inline comments, submitted reviews, and issue comments against the current branch-vs-main diff. The 18 review threads classify as 14 outdated/already-fixed or superseded-by-the-current implementation and 4 current/already-addressed. One previously resolved human finding had regressed: `.gstack/` was removed in response to review, then reintroduced later by `07dbfb878`. I removed that unrelated ignore entry again and replied on the original thread with the commit-level RCA and proof.

The thermo-nuclear pass found one additional real boundary gap: parity validation compared visible text and links but did not inspect Gmail `application/ld+json` actions, so structured-data actions could drift without failing the gate. The validator now parses and exact-compares every JSON-LD document between legacy and modern output. I also removed a redundant CI build step because `validate:parity` already performs the build.

Verification:

- `npm run lint` — passed
- `npm run check` — 0 errors, 0 warnings
- `npm run validate:parity -- --semantic-only` — 13/13 templates; text, links, and JSON-LD exact
- `npm run validate:parity` — 13/13 semantic checks plus all 26 desktop/mobile pixel comparisons exact
- `npm run build-storybook` — passed
- `npm audit --omit=dev` and `npm audit` — 0 vulnerabilities
- `dotnet test tests/Exceptionless.Tests/Exceptionless.Tests.csproj --no-restore -- --filter-class Exceptionless.Tests.Mail.MailerTests` — 27/27 passed
- `git diff --check` — passed
- [replacement CI run](https://github.com/exceptionless/Exceptionless/actions/runs/30210109540) — email templates, client, E2E/Aspire, container build, version, and CLA passed; full API test/coverage result recorded by the run

No remaining actionable review concern or external blocker was found. The complete structural pass found no >1,000-line source file, unsafe boundary/type leak, duplicated canonical helper, non-atomic update, or clear behavior-preserving broad refactor beyond the focused CI simplification above.

niemyjski:
Addressed the invitation feedback and audited every internal email URL against the Svelte UI.

The organization invitation does not provision an account or password-reset token, so a setup-password page would be the wrong contract. A genuinely new address now follows the invitation email directly to `/signup?token=...`; if that address has acquired an account before the invitation is accepted, `/login?token=...` now preserves the token through password and OAuth login and sends it to the API so organization membership is completed. Direct organization links also retain the organization from the URL.

I migrated the remaining legacy email routes to their Svelte equivalents and extended parity validation to discover SvelteKit routes and verify both clean URLs and the current `/next` deployment prefix.

Verification at `73b07ce9b`:
- focused invitation E2E: 2/2 passed (new signup and existing-account login)
- ClientApp build, full `npm run validate`, and focused auth unit tests passed
- email lint/check passed
- all 13 semantic parity scenarios passed
- all 26 desktop/mobile comparisons are exact pixel matches
- every internal email link maps to a real Svelte route under `/next`
- email package audit: 0 vulnerabilities

Dependencies were updated to the latest compatible releases; remaining outdated entries are major-version upgrades and were intentionally left for separate compatibility work.

niemyjski:
Replacement CI is fully green at `73b07ce9b`: API/coverage, client, email templates, full E2E, container build, version, and CLA all passed. Run: https://github.com/exceptionless/Exceptionless/actions/runs/30580372048

src/Exceptionless.Web/ClientApp/src/lib/features/auth/index.svelte.ts
```diff
@@ -158,6 +163,7 @@ async function oauthLogin(options: OAuthLoginOptions) {
     const response = await client.postJSON<TokenResult>(`auth/${options.provider}`, {
         clientId: options.clientId,
         code: data.code,
+        inviteToken: options.inviteToken,
```
niemyjski: **[P1] Honor invitation tokens before blocking OAuth account creation**

This new client path forwards the token, but a brand-new OAuth identity still cannot use it when account creation is disabled. `ExternalLoginAsync` calls `FromExternalLoginAsync`, whose `user is null` branch checks only `authOptions.EnableAccountCreation`; it returns 403 before `AddInvitedUserToOrganizationAsync` can consume `authInfo.InviteToken`. Password signup already gates through `IsAccountCreationEnabledAsync(model.InviteToken)`, so invited users see OAuth signup buttons that fail in the deployment mode invitations are specifically meant to support. Please make the external-login creation gate token-aware and add an `AuthEndpointTests` case for disabled account creation + a valid invite + a new provider user.

src/Exceptionless.EmailTemplates/src/build-emails.ts
```diff
@@ -0,0 +1,137 @@
...
+import OrganizationAdded from './templates/organization-added.svelte';
+import OrganizationInvited from './templates/organization-invited.svelte';
+import OrganizationNotice from './templates/organization-notice.svelte';
+import OrganizationPaymentFailed from './templates/organization-payment-failed.svelte';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+const templates: Record<string, Component> = {
+    'user-password-reset': UserPasswordReset as unknown as Component,
```
niemyjski: **[P2] Keep the registry’s actual component types**

Every template is erased through `unknown` just to satisfy this `Record<string, Component>` annotation. That disables checking at the canonical registry boundary and repeats the same cast eight times, while `Renderer.render` accepts the imported component values directly. Please let the object infer its value types (and remove the `Component` import/casts), or introduce a real renderable-component contract if the renderer gains one, so an accidental non-component entry is caught rather than asserted away.

niemyjski: Fixed in c521f42f7. The template registry now stays fully inferred with no Component casts, and the build enforces an exact filename match across the registry, all 9 Svelte sources, and all 9 production HTML outputs.

src/Exceptionless.Web/ClientApp/src/lib/features/auth/index.svelte.ts
niemyjski: Fixed in c521f42f7. ExternalLoginAsync now carries the invite token into FromExternalLoginAsync, and the account-creation gate calls IsAccountCreationEnabledAsync(inviteToken). Added FacebookAsync_WithValidInviteAndAccountCreationDisabled_CreatesInvitedUser with explicit Arrange/Act/Assert coverage. The project builds locally; the service-backed integration run is left to hosted CI because local Docker was unhealthy.

Issue #314, opened 2017-07-29
Missing '@version' information in data returned by Exceptionless


CyberSinh:
The data returned by Exceptionless is sometimes inconsistent. 
In most cases, I have the information of the version of my software in the data:

`data ": {" @version ":" 5.7.2.5 "}`

 but more and more often this information is not not present.

Did I miss something? Is there anything I can do to make the return of my software version systematic?

Thank you.

CyberSinh:
Oops.. Issue posted in the Exceptionless.Net thread.

Issue #313, opened 2017-07-18
url to see latest events of a specific user


john-roland:
Hello team,

Is there such url we can forge to see events filtered by a specific user ?

if not, i will go through the API.

Thanks

niemyjski:
Currently no, you'd need to input your search filter `user:user@user.com` into the search box. We'd be more than happy to accept a pr that allows a search filter to be passed into the url.

ejsmith:
We're closing this as part of an open issue cleanup. The latest Svelte UI is URL/filter driven, and user-specific event views can now be represented with filter query state rather than needing a special route.

If a specific user-filter URL is still hard to discover or missing from a particular screen, please open a focused docs or UX issue for that exact gap.

PR #2267, opened 2026-05-29
fix: Optimize cleanup jobs, eliminate raw ES client usage, add test coverage

niemyjski:
## Summary

Cleanup now pages through distinct event identifiers through repository methods, avoiding truncated orphan scans. Strict real-time parent reads fail closed on Elasticsearch errors before deletion, and paged cleanup renews its job lock.

Duplicate stacks retain redirects for late events and buffered counters. An occurrence ledger makes retries idempotent; optimistic concurrency and targeted patches preserve concurrent metadata updates. Internal reconciliation fields stay out of API JSON. Existing routes and response contracts are unchanged.

## Validation

Latest `main` merged without conflicts. Backend build passed with zero warnings/errors; 345 serialization and public compatibility tests passed. [Exact-revision hosted checks](https://github.com/exceptionless/Exceptionless/actions/runs/35682300489) are pending. Local integration tests and dogfood are pending exclusive access to shared test infrastructure.

Parent validation and event deletion remain separate cross-index operations and rely on normal parent-before-event ingestion. Performance parity has not been benchmarked.

## Breaking changes

None.


tests/Exceptionless.Tests/Jobs/CleanupOrphanedDataJobTests.cs
```diff
@@ -0,0 +1,236 @@
...
+using Exceptionless.Tests.Utility;
+using Foundatio.Repositories;
+using Foundatio.Repositories.Utility;
+using Foundatio.Utility;
+using Nest;
+using Xunit;
+
+namespace Exceptionless.Tests.Jobs;
+
+public class CleanupOrphanedDataJobTests : IntegrationTestsBase
```
niemyjski, line 15: tests need 3 name part with act, assert, arrange. check pr.

tests/Exceptionless.Tests/Jobs/CleanupDataJobTests.cs
```diff
@@ -169,4 +170,152 @@ public async Task CanDeleteOrphanedEventsByStack()
         eventCount = await _eventRepository.CountAsync(o => o.IncludeSoftDeletes().ImmediateConsistency());
         Assert.Equal(5000, eventCount);
     }
+
+    [Fact]
+    public async Task CanCleanupMultipleSoftDeletedOrganizations()
```
niemyjski: tests need 3 name part with act, assert, arrange. check pr.

src/Exceptionless.Core/Repositories/StackRepository.cs
```diff
@@ -188,6 +188,18 @@ public Task<long> SoftDeleteByProjectIdAsync(string organizationId, string proje
         );
     }
 
+    public async Task<IReadOnlyCollection<string>> GetDuplicateSignaturesAsync(int maxResults = 10000)
```
niemyjski, line 246: Make sure we have integration tests for this.

src/Exceptionless.Core/Repositories/StackRepository.cs
```diff
@@ -188,6 +188,18 @@ public Task<long> SoftDeleteByProjectIdAsync(string organizationId, string proje
...
+    {
+        var result = await CountAsync(q => q.FilterExpression("is_deleted:false")
```
niemyjski: soft deletes should be filtered out by default.

src/Exceptionless.Core/Repositories/StackRepository.cs
```diff
@@ -188,6 +188,18 @@ public Task<long> SoftDeleteByProjectIdAsync(string organizationId, string proje
...
+            .AggregationsExpression($"terms:(duplicate_signature~{maxResults} @min_count:2)"));
+
+        var buckets = result.Aggregations.Terms("terms_duplicate_signature")?.Buckets;
+        if (buckets == null || buckets.Count == 0)
+            return Array.Empty<string>();
```
niemyjski: use pattern matching and use []

src/Exceptionless.Core/Repositories/EventRepository.cs
```diff
@@ -194,4 +197,81 @@ public Task<long> RemoveAllByStackIdsAsync(string[] stackIds)
 
         return RemoveAllAsync(q => q.Stack(stackIds));
     }
+
+    public Task<long> RemoveAllByProjectIdsAsync(string[] projectIds)
+    {
+        ArgumentNullException.ThrowIfNull(projectIds);
+        if (projectIds.Length == 0)
```
niemyjski: prefer is and is not over == and !=, check pr.

src/Exceptionless.Core/Repositories/EventRepository.cs
```diff
@@ -194,4 +197,81 @@ public Task<long> RemoveAllByStackIdsAsync(string[] stackIds)
...
+    public Task<long> RemoveAllByOrganizationIdsAsync(string[] organizationIds)
+    {
+        ArgumentNullException.ThrowIfNull(organizationIds);
+        if (organizationIds.Length == 0)
+            throw new ArgumentOutOfRangeException(nameof(organizationIds));
+
+        return RemoveAllAsync(q => q.Organization(organizationIds));
+    }
+
+    public Task<long> ReassignStackAsync(IEnumerable<string> sourceStackIds, string targetStackId)
```
niemyjski: this feels dangerous where is it used? make sure we have integration tests for this and every method added in repos.

src/Exceptionless.Core/Repositories/EventRepository.cs
```diff
@@ -194,4 +197,81 @@ public Task<long> RemoveAllByStackIdsAsync(string[] stackIds)
...
+                    if (afterKey?.AfterKey.Count > 0)
+                        composite.After(new CompositeKey(afterKey.AfterKey));
+                    return composite;
+                }));
+            return s;
+        });
+
+        var composite = search.Aggregations.Composite($"composite_{fieldName}");
+        if (composite?.Buckets == null || composite.Buckets.Count == 0)
+            return Array.Empty<string>();
```
niemyjski: pattern matching here and return [] check pr for things like this.

src/Exceptionless.Core/Repositories/EventRepository.cs
```diff
@@ -194,4 +197,81 @@ public Task<long> RemoveAllByStackIdsAsync(string[] stackIds)
...
+    {
+        return GetDistinctFieldValuesAsync("organization_id", e => e.OrganizationId, batchSize, afterKey);
+    }
+
+    private async Task<IReadOnlyCollection<string>> GetDistinctFieldValuesAsync(
+        string fieldName, Expression<Func<PersistentEvent, object>> fieldExpression, int batchSize, CompositeKeyResult? afterKey)
+    {
+        var search = await _configuration.Client.SearchAsync<PersistentEvent>(s =>
+        {
+            s.Size(0).Aggregations(a => a
```
niemyjski: we should document here why we are using a composite agg. and performance / concerns.

src/Exceptionless.Core/Repositories/Interfaces/IEventRepository.cs
```diff
@@ -13,6 +13,17 @@ public interface IEventRepository : IRepositoryOwnedByOrganizationAndProject<Per
...
     Task<long> RemoveAllByStackIdsAsync(string[] stackIds);
+    Task<long> RemoveAllByProjectIdsAsync(string[] projectIds);
+    Task<long> RemoveAllByOrganizationIdsAsync(string[] organizationIds);
+    Task<long> ReassignStackAsync(IEnumerable<string> sourceStackIds, string targetStackId);
+    Task<IReadOnlyCollection<string>> GetDistinctStackIdsAsync(int batchSize, CompositeKeyResult? afterKey = null);
+    Task<IReadOnlyCollection<string>> GetDistinctProjectIdsAsync(int batchSize, CompositeKeyResult? afterKey = null);
+    Task<IReadOnlyCollection<string>> GetDistinctOrganizationIdsAsync(int batchSize, CompositeKeyResult? afterKey = null);
+}
+
+public class CompositeKeyResult
```
niemyjski: use record for this

src/Exceptionless.Core/Jobs/CleanupOrphanedDataJob.cs
```diff
@@ -60,305 +62,209 @@ protected override async Task<JobResult> RunInternalAsync(JobContext context)
...
+                    lastStatus = _timeProvider.GetUtcNow().UtcDateTime;
+                    await RenewLockAsync(context);
+                    _logger.LogInformation("Total={Processed}/{Total} Errors={ErrorCount}", processed, total, error);
+                    await _cacheClient.RemoveByPrefixAsync(nameof(Stack));
+                }
+            }
+            catch (Exception ex)
+            {
+                error++;
+                _logger.LogError(ex, "Error fixing duplicate stack {ProjectId} {SignatureHash}", projectId, signature);
```
niemyjski: look at all log statements here we always include :{Message}

src/Exceptionless.Core/Jobs/CleanupOrphanedDataJob.cs
```diff
@@ -60,305 +62,209 @@ protected override async Task<JobResult> RunInternalAsync(JobContext context)
...
+                targetStack.SnoozeUntilUtc = stacks.Documents.Max(d => d.SnoozeUntilUtc);
+                targetStack.DateFixed = stacks.Documents.Max(d => d.DateFixed);
+                targetStack.TotalOccurrences += duplicateStacks.Sum(d => d.TotalOccurrences);
+                targetStack.Tags.UnionWith(duplicateStacks.SelectMany(d => d.Tags));
+                targetStack.References = stacks.Documents.SelectMany(d => d.References).Distinct().ToList();
+                targetStack.OccurrencesAreCritical = stacks.Documents.Any(d => d.OccurrencesAreCritical);
+
+                duplicateStacks.ForEach(s => s.IsDeleted = true);
+                await _stackRepository.SaveAsync(duplicateStacks);
+                await _stackRepository.SaveAsync(targetStack);
```
niemyjski: why do we call this twice when you can just call [] overload.

src/Exceptionless.Core/Jobs/CleanupOrphanedDataJob.cs
```diff
@@ -1,41 +1,41 @@
...
 using Foundatio.Resilience;
 using Microsoft.Extensions.Diagnostics.HealthChecks;
 using Microsoft.Extensions.Logging;
-using Nest;
-using LogLevel = Microsoft.Extensions.Logging.LogLevel;
 
 namespace Exceptionless.Core.Jobs;
 
 [Job(Description = "Deletes orphaned data.", IsContinuous = false)]
 public class CleanupOrphanedDataJob : JobWithLockBase, IHealthCheck
```
niemyjski, line 16: make sure we have 100% coverage for this before we change this and it's passing in origin/main.

src/Exceptionless.Core/Jobs/CleanupDataJob.cs
```diff
@@ -129,13 +132,16 @@ private async Task CleanupSoftDeletedOrganizationsAsync(JobContext context)
 
     private async Task CleanupSoftDeletedProjectsAsync(JobContext context)
     {
-        var projectResults = await _projectRepository.GetAllAsync(o => o.SoftDeleteMode(SoftDeleteQueryMode.DeletedOnly).SearchAfterPaging().PageLimit(5));
+        var projectResults = await _projectRepository.GetAllAsync(o => o.SoftDeleteMode(SoftDeleteQueryMode.DeletedOnly).SearchAfterPaging().PageLimit(100));
```
niemyjski: should we be changing the page size in this pr? wouldn't it be better to get smaller projects incase they are removed etc.. and search over those.

src/Exceptionless.Core/Jobs/CleanupDataJob.cs
```diff
@@ -129,13 +132,16 @@ private async Task CleanupSoftDeletedOrganizationsAsync(JobContext context)
...
         _logger.CleanupProjectSoftDeletes(projectResults.Total);
 
         while (projectResults.Documents.Count > 0 && !context.CancellationToken.IsCancellationRequested)
```
niemyjski, line 244: during cleanup what's renewing the job lock?

tests/Exceptionless.Tests/Jobs/CleanupOrphanedDataJobTests.cs
niemyjski: Addressed in 92fa434fb. The cleanup/orphan tests now use three-part names describing the method/situation/expected result, and the focused CleanupOrphanedDataJobTests suite passes.

tests/Exceptionless.Tests/Jobs/CleanupDataJobTests.cs
niemyjski: Addressed in 92fa434fb. The CleanupDataJobTests names now follow method/situation/expected-result naming, and the focused CleanupDataJobTests suite passes.

src/Exceptionless.Core/Repositories/StackRepository.cs
niemyjski: Addressed in 92fa434fb. StackRepositoryTests now include integration coverage for duplicate signatures, no duplicate signatures, and soft-deleted stacks.

src/Exceptionless.Core/Repositories/StackRepository.cs
niemyjski: Addressed in 92fa434fb. GetDuplicateSignaturesAsync relies on the repository default soft-delete filtering, and GetDuplicateSignatures_WithSoftDeletedStacks_ExcludesThem verifies soft-deleted stacks are excluded.

src/Exceptionless.Core/Repositories/StackRepository.cs
niemyjski: Addressed in 92fa434fb. The duplicate-signature aggregation path now uses pattern matching for empty buckets and returns an empty collection with the [] expression.

src/Exceptionless.Core/Repositories/EventRepository.cs
niemyjski: Addressed in 92fa434fb. The touched EventRepository paths were cleaned up to use pattern matching and is/is not style where applicable while preserving behavior.

src/Exceptionless.Core/Repositories/EventRepository.cs
niemyjski: Addressed in 92fa434fb. ReassignStackAsync is used by CleanupOrphanedDataJob duplicate-stack repair, now materializes and guards empty source ids, and has integration coverage for source events and empty-source no-op behavior.

src/Exceptionless.Core/Repositories/EventRepository.cs
niemyjski: Addressed in 92fa434fb. The distinct-id aggregation path now uses pattern matching for bucket checks and returns [] for empty results.

src/Exceptionless.Core/Repositories/EventRepository.cs
niemyjski: Addressed in 92fa434fb. GetDistinctFieldValuesAsync includes XML documentation explaining why composite aggregations are used and the sequential pagination/performance tradeoff.

src/Exceptionless.Core/Repositories/Interfaces/IEventRepository.cs
niemyjski: Addressed in 92fa434fb. CompositeKeyResult is now a record and uses the Elastic cursor value types required by the current client.

src/Exceptionless.Core/Jobs/CleanupOrphanedDataJob.cs
niemyjski: Addressed in 92fa434fb. The duplicate-stack error log now includes :{Message} with ex.Message while preserving the exception object.

src/Exceptionless.Core/Jobs/CleanupOrphanedDataJob.cs
niemyjski: Addressed in 92fa434fb. Duplicate stack saves now use the collection overload with the combined duplicate stacks plus target stack.

src/Exceptionless.Core/Jobs/CleanupOrphanedDataJob.cs
niemyjski: Addressed in 92fa434fb. I added focused integration coverage around the cleanup/orphan paths and verified CleanupOrphanedDataJobTests, CleanupDataJobTests, EventRepositoryTests, and StackRepositoryTests against origin/main after the merge.

src/Exceptionless.Core/Jobs/CleanupDataJob.cs
niemyjski: Addressed in 92fa434fb. CleanupDataJob keeps the small paged cleanup shape, and the merge added lock renewal around paged cleanup rather than broadening project batch size.

src/Exceptionless.Core/Jobs/CleanupDataJob.cs
niemyjski: Addressed in 92fa434fb. CleanupDataJob and CleanupOrphanedDataJob renew the job lock during paged cleanup and long-running retention/duplicate-stack loops.

src/Exceptionless.Core/Repositories/EventRepository.cs
```diff
@@ -189,9 +192,98 @@ public override Task<FindResults<PersistentEvent>> GetByProjectIdAsync(string pr
...
+    /// Painless script (no string interpolation) to prevent script injection.
+    /// </summary>
+    public Task<long> ReassignStackAsync(IEnumerable<string> sourceStackIds, string targetStackId)
+    {
+        ArgumentNullException.ThrowIfNull(sourceStackIds);
+        ArgumentException.ThrowIfNullOrEmpty(targetStackId);
+
+        return PatchAllAsync(
+            q => q.Stack(sourceStackIds),
+            new ScriptPatch("ctx._source.stack_id = params.targetStackId")
```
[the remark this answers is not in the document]
niemyjski: Addressed in 92fa434fb. ReassignStackAsync materializes the source ids and returns 0 for an empty source list so it cannot patch every event. EventRepositoryTests covers the empty-source no-op.

src/Exceptionless.Core/Jobs/CleanupOrphanedDataJob.cs
```diff
@@ -60,305 +62,212 @@ protected override async Task<JobResult> RunInternalAsync(JobContext context)
...
 
-        var duplicateStackAgg = await _elasticClient.SearchAsync<Stack>(q => q
-            .QueryOnQueryString("is_deleted:false")
-            .Size(0)
-            .Aggregations(a => a.Terms("stacks", t => t.Field(f => f.DuplicateSignature).MinimumDocumentCount(2).Size(10000))));
-        _logger.LogRequest(duplicateStackAgg, LogLevel.Trace);
-
-        var buckets = duplicateStackAgg.Aggregations.Terms("stacks")?.Buckets ?? new List<Nest.KeyedBucket<string>>();
-        int total = buckets.Count;
+        var duplicateSignatures = await _stackRepository.GetDuplicateSignaturesAsync();
```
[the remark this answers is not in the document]
niemyjski: Addressed in 92fa434fb. FixDuplicateStacks now loops until GetDuplicateSignaturesAsync returns no duplicate signatures, and the repository aggregation forces immediate consistency between batches so soft-deleted duplicates are excluded.

niemyjski:
Merged origin/main and addressed the outstanding review feedback in 92fa434fb.

Summary:
- Removed the accidental packages/ diff from the PR; the final diff against origin/main is back to the intended cleanup/repository/test files.
- Added/kept integration coverage for the new repository methods, duplicate signature lookup, soft-delete filtering, empty ReassignStackAsync safety, orphan cleanup, and cleanup usage accounting.
- Updated duplicate-stack cleanup to loop until no duplicate signatures remain and to renew locks during long-running cleanup paths.
- Documented the composite aggregation cursor path and aligned it with the current Elastic client API.
- Treated missing event indexes as zero matching events during event cleanup, which fixes empty soft-deleted project cleanup.

Verification:
- dotnet test -- --filter-class Exceptionless.Tests.Jobs.CleanupDataJobTests: 26 passed
- dotnet test -- --filter-class Exceptionless.Tests.Jobs.CleanupOrphanedDataJobTests: 24 passed
- dotnet test -- --filter-class Exceptionless.Tests.Repositories.EventRepositoryTests: 17 passed, 1 skipped performance test
- dotnet test -- --filter-class Exceptionless.Tests.Repositories.StackRepositoryTests: 13 passed

All review threads are resolved.

niemyjski:
Production-readiness pass complete on f6fcc4b9524a718fa9f45e1e1d38adca80645b7a.

- Merged current origin/main with normal merge commit 65fe08ef4c7d46c5675fa69ae36167d8fa6eada3.
- Added lossless concurrent cleanup hardening, exact event-index verification/fail-closed behavior, durable/transitive duplicate-stack redirects, bounded late-event reconciliation, optimistic concurrency, and patch-based hot update paths.
- Added integration coverage for closed/missing indexes, partial failure/idempotency, stale saves, chain/cycle redirects, late events and counters, tenant deletion, serialization boundaries, and concurrent tag/status/regression updates.
- Local verification: dotnet build --no-restore -m:1 passed with 0 warnings/errors; dotnet test --no-build passed 2336 tests with 2 intentional performance-test skips and 0 failures.
- Current Build workflow is green: version, test-api with coverage, test-client, test-e2e with localhost AppHost, and docker-build all succeeded. Non-PR deploy/publish/preview jobs were skipped as designed.
- Final GraphQL audit: all 21 review threads are resolved; 0 unresolved actionable threads. PR merge state is CLEAN.

Production-readiness verdict: READY.

niemyjski:
Production-readiness refresh complete on bd560b0072b5f83341e41cd73699ac2f16deab61.

- Merged current origin/main (87af2a020282ab44394c50e4ac01ffeb056a8266) with normal merge commit cfc28b77a; no rebase or force-push.
- Re-reviewed the full 18-file PR diff for correctness, compatibility, security, test coverage, and maintainability.
- Fixed the remaining documented convention issue by renaming FixDuplicateStacksAsync and aligning new test names with MethodUnderTest_Scenario_ExpectedBehavior.
- Updated the PR description to document the durable redirect/reconciliation, optimistic-concurrency, and patch-based hot-path hardening already present in the final diff.
- Local build: dotnet build --disable-build-servers --no-restore -m:1 passed with 0 warnings and 0 errors.
- Local service-free contract proof: PublicApiCompatibilityTests passed 4/4. The service-backed local suite was contaminated by concurrent shared Elasticsearch state (alias/index collisions, duplicate fixed test ids, missing shards); isolated CI is authoritative.
- GitHub Actions run 29467194402 is green: version, test-api with the full .NET suite and coverage, test-client, localhost Aspire/Playwright test-e2e, and docker-build all succeeded. Deploy/publish/preview-only jobs skipped as designed.
- Final GraphQL audit: 21/21 review threads resolved, 0 unresolved; merge state CLEAN; no incoming main/head/review changes.

Production-readiness verdict: READY. PR remains open and unmerged.

niemyjski:
Follow-up reviewer audit complete on c2bcdc623b4f177b9828e34ba0983ed921d207ce.

Feedback ledger
- 21 review-thread roots and 38 inline comments were re-read against the current 18-file diff. All 21 threads were already resolved and outdated; no open thread required a new reply or resolution.
- Already fixed/outdated in current code and focused coverage: #3321326717, #3327871089, #3327871203, #3327871916, #3327872423, #3327872834, #3327873570, #3327875527, #3327876881, #3327877605, #3327877922, #3327878657, #3327879396, #3327880341, #3327881304, #3327881626, #3327927526, and #3327927540. These cover the obsolete cursor null-dereference shape, test naming/AAA, repository integration coverage, soft-delete filtering, pattern style, composite pagination documentation and record shape, structured logging, batch save/lock renewal, empty-source safety, and duplicate aggregation pagination.
- Stale/incorrect for the current contract: #3321326725 and #3327911191 duplicate generic-catch findings. Cooperative cancellation is explicitly rethrown; the remaining per-signature catch intentionally isolates corrupt, repository, transport, and transient failures so one duplicate group does not abort the whole cleanup job. Restricting it to TimeoutException and InvalidOperationException would regress that contract.
- Superseded by diff cleanup: #3329055683 targeted generated packages/ content that is no longer in the PR.
- Submitted reviews: Copilot review #4394198870 is an informational overview whose two findings are fixed above; #4397767971 was unable to review temporary 300-plus-file generated churn and is superseded by the current 18-file diff. The remaining submitted reviews have empty bodies and only contain the inline comments classified above.
- Top-level comments #4742267061, #4952560180, and #4987690483 are historical readiness summaries; #5081447901 is the current informational 77 percent coverage report. None contains an unresolved action.

RCA and current design proof
- Empty source ids previously omitted the stack filter and could match every event; ReassignStackAsync now materializes and rejects an empty source set, with regression coverage.
- A single duplicate-signature terms page could strand groups above the cap; cleanup now refreshes and loops until no duplicate buckets remain.
- Concurrent and in-flight updates could make duplicate cleanup lossy; durable redirects, strict reassignment verification, an idempotent occurrence ledger, reconciliation markers, optimistic concurrency, and patch-based hot updates preserve late events and metadata.
- Null duplicate metadata could be encoded as DateTime.MinValue and persisted as fake dates; the current head uses explicit presence flags and regression assertions.

Thermo-nuclear review
- Reviewed the complete branch-vs-main diff for code-judo opportunities, branching growth, boundaries, canonical helpers, sequential orchestration, and partial updates. No production file crosses 1,000 lines; the largest is 568 lines.
- Raw Elasticsearch operations are confined to the EventRepository boundary for composite pagination and strict closed-index verification. Cleanup jobs use repository abstractions.
- The serialized tag-patch loop is intentional cluster backpressure for batches of up to 1,000 distinct stacks; unbounded parallelization would create a new Elasticsearch burst rather than a safe simplification.
- No proven structural or correctness defect remains, so this audit made no speculative code change.

Verification
- dotnet build -m:1: passed, 0 warnings, 0 errors.
- PublicApiCompatibilityTests: 4/4 passed.
- StackRepositoryTests local attempt: 24 fixture-initialization errors before assertions because concurrent local hosts deleted and recreated the shared test-* indices, yielding index_not_found_exception and invalid_alias_name_exception. This is isolated as shared infrastructure state, not a branch failure.
- Authoritative clean run 30183197271 for this exact SHA: version, test-api with full .NET tests and coverage, test-client, localhost Aspire/Playwright test-e2e, and docker-build all passed.
- git diff --check passed; live PR state was OPEN, MERGEABLE, and CLEAN.

Verdict: READY. No code changes were required in this follow-up audit; the PR remains open and unmerged.

src/Exceptionless.Core/Jobs/CleanupOrphanedDataJob.cs
```diff
@@ -58,211 +50,190 @@ ILoggerFactory loggerFactory
...
         {
             await RenewLockAsync(context);
 
-            var stackIdTerms = await _elasticClient.SearchAsync<PersistentEvent>(s => s
-                .Indices(GetEventIndexPattern())
-                .Size(0)
-                .AddAggregation("terms_stack_id", a => a.Terms(c => c.Field(f => f.StackId).Include(new TermsInclude(batchNumber, buckets)).Size(batchSize * 2))));
+            var page = await _eventRepository.GetDistinctStackIdsAsync(500, nextValue, context.CancellationToken);
+            var stackIds = page.Values;
+            if (stackIds.Count == 0)
```
niemyjski: prefer is 0 always

src/Exceptionless.Core/Jobs/CleanupOrphanedDataJob.cs
```diff
@@ -58,211 +50,190 @@ ILoggerFactory loggerFactory
...
+            nextValue = page.NextValue;
+            hasMore = !String.IsNullOrEmpty(nextValue);
+            totalStackIds += stackIds.Count;
 
-            totalStackIds += stackIds.Length;
-
-            var stacks = await _stackRepository.GetByIdsAsync(stackIds, o => o.ImmediateConsistency());
-            var foundStackIds = stacks.Select(stack => stack.Id).ToHashSet(StringComparer.OrdinalIgnoreCase);
-            string[] missingStackIds = stackIds.Where(stackId => !foundStackIds.Contains(stackId)).ToArray();
+            var existingStacks = await _stackRepository.GetByIdsAsync(stackIds.ToArray(), o => o.Include(s => s.Id, s => s.IsDeleted));
```
niemyjski: is this just doing a cheap exists? soft deleted won't be returned by default so this feels like an onlyIds?

src/Exceptionless.Core/Repositories/EventRepository.cs
```diff
@@ -188,12 +197,244 @@ public override Task<FindResults<PersistentEvent>> GetByProjectIdAsync(string pr
...
+        }
+        catch (RepositoryException ex) when (IsIndexNotFound(ex.InnerException as TransportException))
+        {
+            return 0;
+        }
+        catch (TransportException ex) when (IsIndexNotFound(ex))
+        {
+            return 0;
+        }
+    }
```
niemyjski: why do we have all these one off methods? Worst of all they really don't do anything that the built in ones do.

src/Exceptionless.Core/Repositories/StackRepository.cs
```diff
@@ -72,7 +100,10 @@ Instant parseDate(def dt) {
   ctx._source.updated_utc = params.updatedUtc;
 }
 
-ctx._source.total_occurrences += params.count;";
+ctx._source.total_occurrences += params.count;
+if (ctx._source.redirect_to_stack_id != null) {
+  ctx._source.needs_redirect_reconciliation = true;
```
niemyjski, line 107: why do we have or need needs_redirect_reconciliation

src/Exceptionless.Core/Jobs/CleanupOrphanedDataJob.cs
niemyjski: Updated every touched C# zero-count check to use `is 0`, including the stack/project/organization cleanup loops, duplicate batches, and repository guards. The remaining `== 0` checks are inside Painless scripts, where C# pattern syntax does not apply. The solution builds with zero warnings/errors.

src/Exceptionless.Core/Jobs/CleanupOrphanedDataJob.cs
niemyjski: I kept this as `GetByIdsAsync(... Include(Id, IsDeleted))` intentionally. This destructive decision needs real-time, fail-closed existence semantics. `OnlyIds` is search-backed; Foundatio logs a refresh failure and can continue with stale results, which could miss a just-written live parent and delete valid events. It also returns IDs through `Hits`, not `Documents`. MGET applies default soft-delete filtering client-side, so `IsDeleted` must be hydrated. I added an unrefreshed-live-stack regression to preserve this invariant.

src/Exceptionless.Core/Repositories/EventRepository.cs
niemyjski: Agreed. I removed the new organization/project overrides, options overload, plural helpers, and missing-index wrapper. The job now calls inherited `RemoveAllAsync(q => ...)` directly; the installed Foundatio implementation already uses `IgnoreUnavailable` and conflict retries. I retained only the pre-existing one-argument `RemoveAllByStackIdsAsync` for public API compatibility.

src/Exceptionless.Core/Repositories/StackRepository.cs
niemyjski: This is required for a buffered-counter race: events can finish reassignment, then a delayed counter/stat-repair write can land on the redirect tombstone. With no source events left, the orphan scan has nothing to rediscover, so the durable dirty bit is the only retry signal. Both counter scripts set it atomically, and reconciliation applies deltas through an idempotent per-source ledger. I documented that invariant and hardened it further: fresh tombstones are re-merged before a CAS clear, the CAS fences total/first/last/timestamp/target, and first occurrence is now merged too. Coverage now includes increment, set, same-timestamp stale-clear, late-counter/no-events, and closed-index recovery paths.

niemyjski:
Final destructive-cleanup and query audit complete on `ba1aca0b547689b9e7a7fbddefb010c6c7cab9e7`, after merging current `main` normally.

### Coverage and confidence

This repository does **not** have 100% coverage, so I cannot honestly claim that no change can ever break anything. The exact-head hosted result is 78% line coverage (23,782/30,603) and 67% branch coverage (11,127/16,628); Core is 75%/67%.

The authoritative isolated run is green: 2,742 .NET tests, 2,740 passed, 0 failed, and 2 intentional performance-test skips. All four new `CleanupOrphanedDataJobSafetyTests` ran and passed. Version, client, localhost Aspire/Playwright E2E, all Docker image builds, and CLA are also green. All 25 review threads are resolved.

### Deletion-safety findings and fixes

- The audit found a real fail-open path in the installed Foundatio multi-get behavior: Elasticsearch can return HTTP 200 with a per-item closed/missing-index error, and the standard helper logs and omits that item. Cleanup could therefore interpret an unreadable parent as missing and delete its events.
- The cleanup paths now use a strict real-time MGET at the repository boundary. It preserves the existing one-request/source-filtered wire shape, but throws on top-level errors, every per-item error, malformed/null/mismatched sources, and omitted/duplicate/unexpected IDs before any delete-by-query can run.
- Three real-Elasticsearch tests close the stack, project, and organization indices independently and prove cleanup fails while preserving the event.
- A second race was fixed: if a soft-deleted stack is restored between the active-parent read and the all-soft-deletes recheck, it is now recognized as active and excluded from deletion. A deterministic `BeforeGet` integration test forces that exact interleaving and proves the stack and event survive.
- `NeedsRedirectReconciliation` remains necessary. Buffered counter/stat writes can land on a redirect tombstone after every source event has moved; with no source event left, the dirty marker is the durable retry signal. The inline reply documents that invariant and its increment/set/CAS/recovery coverage.

One residual limit is explicit: parent validation and event delete-by-query are separate cross-index operations. Supported ingestion writes the parent before the event, which satisfies the runtime invariant, but an out-of-contract event-first import or concurrent parent restore after the final check can still race cleanup. Eliminating that class absolutely would require an age grace period, two-pass quarantine, or cross-index coordination. Green tests cannot turn that non-atomic design into a mathematical guarantee.

### Query/performance verdict

The result is mixed, not uniformly “faster”:

- Orphan ID discovery is more exact and scalable: bounded composite `after_key` pages replace approximate cardinality plus partitioned terms scans. It removes the initial cardinality request and cannot truncate/skew IDs, but end-to-end latency has not been benchmarked.
- Parent checks still issue one real-time, source-filtered MGET per batch; strict validation adds only O(n) client-side checks over at most 500 items. The stack-miss path intentionally performs a second MGET to protect redirect tombstones and concurrently restored stacks.
- Normal orphan deletion remains the same terms delete-by-query shape; Foundatio adds unavailable-index handling and conflict retries.
- Duplicate reassignment/reconciliation and per-stack tag patching intentionally add reads/writes to prevent stale overwrites, missed conflicts, and data loss. These are safety costs, not performance optimizations. The many-unique-tagged-stacks pipeline is the highest-priority latency benchmark.

CI proves behavior, not performance parity. A defensible performance gate would compare `main` and this head against cloned identical Elasticsearch/Redis snapshots, with warmed runs at 499/500/501/10k/50k IDs and duplicate/tag workloads, capturing request counts, bytes, Elasticsearch `took`, conflicts/batches/throttling, and median/p95 wall time under an explicit regression budget.

Verdict: **merge-ready for the supported parent-before-event runtime invariant**, with substantially stronger fail-closed behavior and exact-head integration proof. It is not 100%-covered, not an absolute cross-index no-loss guarantee, and not yet performance-benchmarked.

niemyjski:
Upstream follow-up for the strict MGET safety finding:

- Filed [Foundatio.Repositories #313](https://github.com/FoundatioFx/Foundatio.Repositories/issues/313) with the HTTP-200/per-item-error reproduction, partial-result/cache-poisoning impact, and missing coverage.
- Opened [Foundatio.Repositories PR #314](https://github.com/FoundatioFx/Foundatio.Repositories/pull/314), adding an opt-in `ThrowOnMultiGetErrors()` command option. It preserves the existing default for compatibility, throws before returning/caching partial results when enabled, and covers both a closed index and a mixed time-series success/error response.
- Upstream local validation: full solution build passed with 0 warnings/errors; 841/841 tests passed; formatting verification passed. Hosted upstream checks are running.

Exceptionless should retain its current fail-closed adapter until a released Foundatio.Repositories package contains #314. After upgrading, the intended cleanup is to replace the duplicated raw MGET implementation and internal interface-to-base-class cast with canonical `GetByIdsAsync(...ThrowOnMultiGetErrors())` calls while keeping the stack/project/organization outage-preservation tests.

I also merged the latest `origin/main` into this PR normally (`283d7ee89`); those five incoming commits are frontend-only. Fresh exact-head Exceptionless CI is running.

niemyjski:
Final exact-head thermo and strict-MGET follow-up:

- Current head `283d7ee892089bcbf2722581a5d58b9dcdcae0dd` includes current `main`. [Build run 31554231590](https://github.com/exceptionless/Exceptionless/actions/runs/31554231590) is fully green: version, API/.NET, client, E2E, Docker, and CLA.
- .NET: 2,742 total; 2,740 succeeded; 0 failed; 2 skipped. The only skips are the unrelated performance-data generators. All four `CleanupOrphanedDataJobSafetyTests` therefore ran and passed in the unfiltered isolated suite.
- Coverage is 78% line / 67% branch overall (Core 75% / 67%). This is not 100% coverage, so I am not claiming mathematical zero-risk; the confidence comes from the targeted destructive-path, outage, pagination, redirect, race, recovery, and idempotency tests plus the full suite.
- All 25 live review threads are resolved. The PR is CLEAN and MERGEABLE.

The MGET behavior was a real upstream Foundatio.Repositories gap: HTTP-200 per-item errors were logged and omitted, making them indistinguishable from missing documents. I filed [issue #313](https://github.com/FoundatioFx/Foundatio.Repositories/issues/313) and opened [PR #314](https://github.com/FoundatioFx/Foundatio.Repositories/pull/314). Both hosted build runs pass 841/841 tests with 0 failures/skips; CodeQL and CLA are green. The fix is opt-in/backward-compatible and adds no request round trip: strict mode changes only per-item response validation.

Thermo verdict: no remaining blocking finding in this PR. Keep the local fail-closed adapter until #314 is released and the package is upgraded; then replace it with canonical `GetByIdsAsync(...ThrowOnMultiGetErrors())` while retaining the four application-level outage/race tests. The remaining parent-check/delete TOCTOU is a pre-existing cross-index architectural boundary; normal ingestion preserves parent-before-event ordering. Supporting event-first imports or concurrent restores with an absolute no-overdelete guarantee would require a separate grace/two-pass/quarantine design.

PR #2268, opened 2026-05-29
feat: add Deleted counter to usage metrics

niemyjski:
## Why

Until now there was no way to see how many events were explicitly deleted via the API, project reset, or soft-delete cleanup. This makes it impossible to distinguish "events never arrived" from "events were intentionally removed" when analyzing usage. This PR adds a first-class `Deleted` counter alongside the existing `Total`, `Blocked`, `Discarded`, and `TooBig` counters -- without touching billing, plan limits, or admission control.

### Backend

- **`UsageInfo` / `UsageHourInfo`** -- added `int Deleted` property to both records, matching the type of all other usage counters (`Total`, `Blocked`, `Discarded`, `TooBig`).
- **`AppDiagnostics`** -- added `EventsDeleted` as a `Counter<int>` OTel metric (consistent with all other event counters).
- **`UsageService`** -- new `IncrementDeletedAsync(organizationId, projectId, eventCount)` that pipelines all Redis operations via `Task.WhenAll`; cache key pattern: `usage:{bucket}:{orgId}[:projectId]:deleted`. Save/flush/read logic updated to include the Deleted bucket (mirrors the other counters exactly).
- **`LastEventDateUtc` behavior change** -- `LastEventDateUtc` on org and project is now only updated when a bucket contains actual ingestion events (`Total`/`Blocked`/`Discarded`/`TooBig` > 0). Previously any flush (including delete-only buckets) could bump this timestamp. Deletes no longer count as "activity" for inactivity tracking.
- **`EventController`** -- `DeleteModelsAsync` calls `IncrementDeletedAsync` per org/project group after the hard delete completes, inside a try/catch that rethrows `OperationCanceledException` and warns on all other failures so a tracking error never blocks the delete response.
- **`ResetProjectDataWorkItemHandler`** -- calls `IncrementDeletedAsync` with the count returned by the delete query.
- **`CleanupDataJob`** -- `RemoveProjectsAsync` tracks deleted events; `RemoveStacksAsync` accepts a `trackDeletedUsage` flag. When `false` (retention enforcement), a single bulk `RemoveAllByStackIdsAsync` is used. When `true` (soft-delete cleanup), the stacks are grouped by `(OrgId, ProjectId)` and deleted per group so per-project counts can be attributed. `RemoveOrganizationAsync` intentionally does NOT track -- the org is being hard-deleted so the data has no consumer.

### Frontend (Svelte 5)

- Generated types (`api.ts`, `schemas.ts`) updated to include `deleted: number`.
- Org and project usage pages: `Deleted` series added to charts, dead variable alias removed.

### Frontend (Angular legacy)

- Org and project manage controllers: `Deleted` series added at the correct index.

## Test coverage

| Test class | New tests |
|---|---|
| `UsageServiceTests` | 6 new (CanIncrementDeletedAsync, flush, read-back, isolation, zero-guard, zero-event-count guard) |
| `EventControllerTests` | 2 integration tests: single delete and multi-event delete with org+project flush assertion |
| `CleanupDataJobTests` | 5 tests: soft-deleted project tracks usage, empty project no-op, soft-deleted stack tracks usage, multi-project distribution, retention enforcement does NOT track |
| `ResetProjectDataWorkItemHandlerTests` | 5 tests: basic tracking, empty project, multiple stacks, project isolation, no billing impact |

Existing tests updated to assert `Deleted == 0` where appropriate.

## Key design decisions

- **Retention deletions are NOT tracked.** `RemoveStacksAsync` is called with `trackDeletedUsage=false` for retention enforcement paths. Only explicit user-initiated deletes and soft-delete cleanup show up as `Deleted` usage.
- **`int` not `long`.** Matches all other usage counters. A single cleanup pass removing >2.1B events per project is not a realistic scenario; the `eventCount <= 0` guard in `IncrementDeletedAsync` safely drops any overflow.
- **Per-project deletion counts are exact**, not estimated -- each group's count comes directly from the ES delete-by-query return value.
- **No billing changes.** `Deleted` is a display-only counter. It does not affect `GetEventsLeftAsync`, plan enforcement, or Stripe metering.

### DIFF
--- a/src/Exceptionless.Core/Jobs/CleanupDataJob.cs
+++ b/src/Exceptionless.Core/Jobs/CleanupDataJob.cs
@@ -226,14 +226,22 @@ public class CleanupDataJob : JobWithLockBase, IHealthCheck
         var projectGroups = stacks.GroupBy(s => (s.OrganizationId, s.ProjectId)).ToList();
 
         long totalRemovedEvents = 0;
-        foreach (var projectGroup in projectGroups)
+        if (trackDeletedUsage)
         {
-            string[] stackIds = projectGroup.Select(s => s.Id).ToArray();
-            long removedEvents = await _eventRepository.RemoveAllByStackIdsAsync(stackIds);
-            totalRemovedEvents += removedEvents;
+            foreach (var projectGroup in projectGroups)
+            {
+                string[] stackIds = projectGroup.Select(s => s.Id).ToArray();
+                long removedEvents = await _eventRepository.RemoveAllByStackIdsAsync(stackIds);
+                totalRemovedEvents += removedEvents;
 
-            if (trackDeletedUsage && removedEvents > 0)
-                await _usageService.IncrementDeletedAsync(projectGroup.Key.OrganizationId, projectGroup.Key.ProjectId, (int)removedEvents);
+                if (removedEvents > 0)
+                    await _usageService.IncrementDeletedAsync(projectGroup.Key.OrganizationId, projectGroup.Key.ProjectId, (int)removedEvents);
+            }
+        }
+        else
+        {
+            string[] allStackIds = stacks.Select(s => s.Id).ToArray();
+            totalRemovedEvents = await _eventRepository.RemoveAllByStackIdsAsync(allStackIds);
         }
 
         await _stackRepository.RemoveAsync(stacks);
--- a/src/Exceptionless.Core/Jobs/CleanupDataJob.cs
+++ b/src/Exceptionless.Core/Jobs/CleanupDataJob.cs
@@ -27,6 +27,7 @@ public class CleanupDataJob : JobWithLockBase, IHealthCheck
     private readonly ITokenRepository _tokenRepository;
     private readonly IWebHookRepository _webHookRepository;
     private readonly BillingManager _billingManager;
+    private readonly UsageService _usageService;
     private readonly AppOptions _appOptions;
     private readonly ILockProvider _lockProvider;
     private readonly ICacheClient _cacheClient;
@@ -43,6 +44,7 @@ public class CleanupDataJob : JobWithLockBase, IHealthCheck
         ILockProvider lockProvider,
         ICacheClient cacheClient,
         BillingManager billingManager,
+        UsageService usageService,
         AppOptions appOptions,
         TimeProvider timeProvider,
         IResiliencePolicyProvider resiliencePolicyProvider,
@@ -57,6 +59,7 @@ public class CleanupDataJob : JobWithLockBase, IHealthCheck
         _tokenRepository = tokenRepository;
         _webHookRepository = webHookRepository;
         _billingManager = billingManager;
+        _usageService = usageService;
         _appOptions = appOptions;
         _lockProvider = lockProvider;
         _cacheClient = cacheClient;
@@ -164,7 +167,7 @@ public class CleanupDataJob : JobWithLockBase, IHealthCheck
         {
             try
             {
-                await RemoveStacksAsync(stackResults.Documents, context);
+                await RemoveStacksAsync(stackResults.Documents, context, trackDeletedUsage: true);
             }
             catch (Exception ex)
             {
@@ -206,6 +209,9 @@ public class CleanupDataJob : JobWithLockBase, IHealthCheck
         await RenewLockAsync(context);
         long removedEvents = await _eventRepository.RemoveAllByProjectIdAsync(project.OrganizationId, project.Id);
 
+        if (removedEvents > 0)
+            await _usageService.IncrementDeletedAsync(project.OrganizationId, project.Id, (int)removedEvents);
+
         await RenewLockAsync(context);
         long removedStacks = await _stackRepository.RemoveAllByProjectIdAsync(project.OrganizationId, project.Id);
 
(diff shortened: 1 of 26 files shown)


src/Exceptionless.Web/ClientApp/src/lib/features/auth/api.svelte.ts
```diff
@@ -97,6 +98,7 @@ export async function login(email: string, password: string) {
 
 export async function logout(queryClient?: QueryClient, client = useFetchClient()) {
     await client.get('auth/logout', { expectedStatusCodes: [200, 401, 403] });
+    endSession().catch(() => {});
```
niemyjski: this is a hack, we should await this, no catch

src/Exceptionless.Web/ClientApp/src/lib/features/billing/components/change-plan-dialog.svelte
```diff
@@ -256,6 +257,7 @@
                     }
 
                     toast.success(result.message ?? 'Your billing plan has been successfully changed.');
+                    submitFeatureUsage('organization.ChangePlan').catch(() => {});
```
niemyjski: I do not like this catch, we should await this, this will never throw.

src/Exceptionless.Web/ClientApp/src/lib/features/users/api.svelte.ts
```diff
@@ -68,6 +69,7 @@ export function getMeQuery() {
         enabled: () => !!accessToken.current,
         onSuccess: (data: ViewCurrentUser) => {
             queryClient.setQueryData(queryKeys.id(data.id!), data);
+            setUserIdentity(data.id, data.full_name).catch(() => {});
```
niemyjski: I do not like this catch, we should await this, this will never throw.

src/Exceptionless.Web/Controllers/EventController.cs
```diff
@@ -1456,16 +1459,31 @@ private async Task<Dictionary<string, double>> GetUserCountByProjectIdsAsync(ICo
...
         }
 
-        return base.DeleteModelsAsync(events);
+        var result = await base.DeleteModelsAsync(events);
+
+        foreach (var projectGroup in projectGroups)
+        {
+            try
+            {
+                await _usageService.IncrementDeletedAsync(projectGroup.Key.OrganizationId, projectGroup.Key.ProjectId, projectGroup.Count());
```
niemyjski, line 1479: would this be double counted by the cleanup job?

src/Exceptionless.Web/Controllers/EventController.cs
```diff
@@ -1456,16 +1459,31 @@ private async Task<Dictionary<string, double>> GetUserCountByProjectIdsAsync(ICo
...
+            }
+            catch (Exception ex)
+            {
+                _logger.LogWarning(ex, "Failed to increment deleted usage metrics for org {OrganizationId} project {ProjectId}", projectGroup.Key.OrganizationId, projectGroup.Key.ProjectId);
```
niemyjski: this should include :{Message}

src/Exceptionless.Web/ClientApp/src/lib/features/auth/api.svelte.ts
niemyjski: Fixed — removed the `.catch(() => {})` hack and now properly awaiting `endSession()`.

src/Exceptionless.Web/ClientApp/src/lib/features/users/api.svelte.ts
niemyjski: Fixed — removed the catch, now properly awaiting `setUserIdentity`. Also made the `onSuccess` callback `async` so the `await` is valid.

src/Exceptionless.Web/Controllers/EventController.cs
niemyjski: Fixed — added `{Message}` to the log template.

src/Exceptionless.Web/ClientApp/src/lib/features/billing/components/change-plan-dialog.svelte
niemyjski: Fixed — removed the catch, now properly awaiting `submitFeatureUsage`.

src/Exceptionless.Web/Controllers/EventController.cs
niemyjski: No — no double counting. The EventController path hard-deletes specific events by ID (events don't implement `ISupportSoftDeletes`). The CleanupDataJob only tracks deletes for soft-deleted *stacks* (`CleanupSoftDeletedStacksAsync`). These are different deletion sources:

- **EventController**: user explicitly deletes specific events via API → count tracked here
- **CleanupDataJob**: stacks soft-deleted (e.g. via StackController) → cleanup physically removes their events → count tracked there

If a user deletes events via API first, then the stack is later soft-deleted and cleaned up, the cleanup would find 0 remaining events for those stacks (`removedEvents == 0`), so `IncrementDeletedAsync` won't fire. No overlap.

src/Exceptionless.Core/Models/UsageInfo.cs
```diff
@@ -18,6 +19,7 @@ public record UsageHourInfo
     public int Blocked { get; set; }
     public int Discarded { get; set; }
     public int TooBig { get; set; }
+    public int Deleted { get; set; }
```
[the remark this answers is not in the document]
niemyjski: Resolved — `int` is the correct type here, matching all other usage counters. PR description was misleading.

src/Exceptionless.Core/Services/UsageService.cs
```diff
@@ -473,6 +489,29 @@ public async Task IncrementTooBigAsync(string organizationId, string? projectId)
         AppDiagnostics.PostTooBig.Add(1);
     }
 
+    public async Task IncrementDeletedAsync(string organizationId, string? projectId, int eventCount = 1)
```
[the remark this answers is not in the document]
niemyjski: Resolved — `int` is intentional and matches all other usage counters (`Total`, `Blocked`, `Discarded`, `TooBig`). The PR description was inaccurate. Also fixed `EventsDeleted` OTel counter from `Counter<long>` to `Counter<int>` for consistency.

src/Exceptionless.Core/Jobs/CleanupDataJob.cs
```diff
@@ -206,24 +209,39 @@ private async Task RemoveProjectsAsync(Project project, JobContext context)
...
+        long totalRemovedEvents = 0;
+        foreach (var projectGroup in projectGroups)
+        {
+            string[] stackIds = projectGroup.Select(s => s.Id).ToArray();
+            long removedEvents = await _eventRepository.RemoveAllByStackIdsAsync(stackIds);
+            totalRemovedEvents += removedEvents;
+
+            if (trackDeletedUsage && removedEvents > 0)
+                await _usageService.IncrementDeletedAsync(projectGroup.Key.OrganizationId, projectGroup.Key.ProjectId, (int)removedEvents);
+        }
```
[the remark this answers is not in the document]
niemyjski: Resolved — the PR description was inaccurate about proportional distribution. The actual implementation groups by (OrgId, ProjectId) and attributes exact per-group counts. No remainder logic is needed because each group's delete count comes directly from ES.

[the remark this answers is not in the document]
niemyjski: Fixed — `RemoveStacksAsync` now uses a single batch `RemoveAllByStackIdsAsync` call when `trackDeletedUsage: false` (retention path). Only splits per-project group when tracking is enabled.

src/Exceptionless.Core/Jobs/WorkItemHandlers/ResetProjectDataWorkItemHandler.cs
```diff
@@ -41,6 +44,9 @@ public override async Task HandleItemAsync(WorkItemContext context)
             long removedEvents = await _eventRepository.RemoveAllByProjectIdAsync(workItem.OrganizationId, workItem.ProjectId);
             await context.ReportProgressAsync(50, $"Events removed: {removedEvents}");
 
+            if (removedEvents > 0)
+                await _usageService.IncrementDeletedAsync(workItem.OrganizationId, workItem.ProjectId, (int)removedEvents);
```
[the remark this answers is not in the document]
niemyjski: Acknowledged — the `(int)` cast is intentional. `int` matches all other usage counters. In practice, a single cleanup pass deleting >2.1B events per project is unrealistic for this system's workload. The guard `eventCount <= 0` correctly handles the theoretical negative-after-overflow case by dropping the count rather than recording garbage.

PR #2284, opened 2026-06-02
Fix UI list filters, stack navigation, and pagination

ejsmith:
## What changed
- Added shared Events/Stacks navigation helpers that map registered filter controls to direct query params like `stack=`, `status=`, `type=`, and `project=`; unmapped filters fall back to raw `filter=` expressions.
- Fixed saved-view URL normalization so saved filters stay out of the URL unless the user changes or removes them.
- Fixed stack-to-events drilldowns to support explicit all-time navigation and cleaner time query params such as `time=1h`.
- Fixed selected stack deletion and stack detail delete navigation so the UI does not remain stuck in loading state.
- Fixed cursor pagination controls so Next/Previous update the table page state and cursor params correctly, with pagination state normalized before applying table updates.
- Persisted Events and Stacks pagination in the URL so copied links reopen to the same list page.
- Omitted `limit` from Events/Stacks URLs when it matches the user's saved rows-per-page preference; explicit `limit` stays when a shared URL overrides that preference.
- Added canonical stack-scoped event detail URLs at `/stack/{stackId}/event/{eventId}` and kept full-page/detail-sheet next/previous navigation in sync with those URLs.
- Prevented duplicate faceted filters, including duplicate Status filters; re-adding a hidden filter now reveals the existing hidden filter instead.
- Fixed system overview all-time chart y-axis clipping.

## Why
User feedback surfaced several high-priority issues in the new Svelte UI around stack drilldowns, saved-view filter URLs, pagination, duplicate filters, stale selection state, and shareable detail URLs. The route contract now uses explicit query params for registered controls instead of serialized filter JSON, while still deriving API filter expressions from the active filter controls.

## Impact
- Users can deep-link to Events or Stacks with exact filter controls.
- Stack-to-events links can intentionally show all-time events without inheriting the current date range.
- Saved-view URLs stay clean and only show deltas from the saved view.
- Event list pagination works with larger page sizes.
- Events and Stacks pagination URLs can be shared and restored, including saved-view list routes.
- Saved rows-per-page preferences keep normal URLs clean while still allowing `limit` overrides in shared links.
- Event detail links include both stack and event ids once the event is loaded, so copied URLs and sheet external-link targets reopen the exact event context.
- Existing hidden filters are surfaced instead of duplicated.

## Breaking changes
None.

## Validation
- `npm run format`
- `npm run lint`
- `npm run check`
- `npm run test:unit -- helpers.svelte redirect-to-events use-saved-views`
- `npm run test:unit -- table.svelte`
- Local browser QA on `https://localhost:7131/next/event/all?limit=50`: verified event pagination Next/Previous and duplicate Status filter prevention, including hidden-filter reveal.
- Local browser QA on `https://localhost:7131/next/event/all?limit=50`: verified Next adds `after` and `page`, reload preserves `Page 2`, and Previous returns to the first-page URL.
- Local browser QA on `https://localhost:7131/next/stack?limit=20`: verified Next adds `page=2` and reload preserves `Page 2`.
- Local browser QA with saved rows-per-page preference set to 50: verified `?limit=50` is removed on Events/Stacks URLs, `?limit=20` remains as an override, and paging adds page/cursor params without re-adding `limit`.
- Local browser QA on stack/event routes: verified time query normalization and all-time stack drilldowns.
- Local browser QA on `https://localhost:7131/next/event/{eventId}`: verified the URL canonicalizes to `/next/stack/{stackId}/event/{eventId}`.
- Local browser QA on full-page stack event details: verified Older/Newer updates the event id in the canonical stack-scoped URL.
- Local browser QA on Events and Stacks detail sheets: verified the "Open details in new window" href uses `/next/stack/{stackId}/event/{eventId}` and follows Older/Newer navigation.

### DIFF
--- a/src/Exceptionless.Web/ClientApp/src/lib/features/shared/table.svelte.ts
+++ b/src/Exceptionless.Web/ClientApp/src/lib/features/shared/table.svelte.ts
@@ -61,6 +61,19 @@ export type TablePagingParameters<T extends PaginationStrategy = PaginationStrat
         ? TableMemoryPagingParameters
         : never;
 
+export function createPageSizePreference(key: string) {
+    const persistedPageSize = new PersistedState<number>(key, DEFAULT_LIMIT);
+
+    return {
+        get current() {
+            return normalizePageSize(persistedPageSize.current);
+        },
+        set current(value: number) {
+            persistedPageSize.current = normalizePageSize(value);
+        }
+    };
+}
+
 export function getSharedTableOptions<TData extends RowData, TPaginationStrategy extends PaginationStrategy = PaginationStrategy>(
     configuration: TableConfiguration<TData, TPaginationStrategy>
 ): TableOptions<StockFeatures, TData> {
@@ -435,6 +448,10 @@ function getPageIndexFromParameters(strategy: PaginationStrategy, parameters: Ta
 
 function hasSortQueryParameter(parameters: TablePagingParameters): parameters is TableCursorPagingParameters | TableOffsetPagingParameters {
     return Object.prototype.hasOwnProperty.call(parameters, 'sort');
+}
+
+function normalizePageSize(value: number | undefined): number {
+    return value && Number.isFinite(value) && value > 0 ? value : DEFAULT_LIMIT;
 }
 
 function parseSortString(sort: string | undefined): ColumnSort[] {
--- a/src/Exceptionless.Web/ClientApp/src/lib/features/shared/table.svelte.ts
+++ b/src/Exceptionless.Web/ClientApp/src/lib/features/shared/table.svelte.ts
@@ -112,12 +112,9 @@ export function getSharedTableOptions<TData extends RowData, TPaginationStrategy
 
     const onPaginationChange = (updaterOrValue: Updater<PaginationState>) => {
         const previousPageInfo = pagination();
-
-        setPagination(updaterOrValue);
-        const paginationChange = getPaginationChange(previousPageInfo, pagination());
-        if (paginationChange.pageIndexChanged) {
-            setPagination(paginationChange.currentPageInfo);
-        }
+        const requestedPageInfo = resolveUpdater(previousPageInfo, updaterOrValue);
+        const paginationChange = resolvePaginationChange(previousPageInfo, requestedPageInfo);
+        setPagination(paginationChange.currentPageInfo);
 
         const currentPageInfo = paginationChange.currentPageInfo;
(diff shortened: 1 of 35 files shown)


src/Exceptionless.Web/ClientApp/src/lib/features/shared/table.svelte.ts
```diff
@@ -111,10 +111,20 @@ export function getSharedTableOptions<TData extends RowData, TPaginationStrategy
...
+        let currentPageInfo = pagination();
+        const pageSizeChanged = previousPageInfo.pageSize !== currentPageInfo.pageSize;
+
+        if (pageSizeChanged && currentPageInfo.pageIndex !== 0) {
+            currentPageInfo = {
+                ...currentPageInfo,
+                pageIndex: 0
+            };
+            setPagination(currentPageInfo);
+        }
```
niemyjski: This feels a bit hacky. this was working. We have three different table pagination strategies.

src/Exceptionless.Web/ClientApp/src/routes/(app)/project/[projectId]/stacks/[stackId]/+page.svelte
```diff
@@ -24,7 +24,8 @@
     );
 
     async function filterChanged(addedOrUpdated: IFilter) {
-        await redirectToEventsWithFilter(organization.current, addedOrUpdated);
+        const options = addedOrUpdated.type === 'string' && addedOrUpdated.key === 'string-stack' ? { time: null } : undefined;
```
Comment (niemyjski) [inline src/Exceptionless.Web/ClientApp/src/routes/(app)/project/[projectId]/stacks/[stackId]/+page.svelte]: can this be a function that's documented, a lot of this stuff was hard to track down.

Comment (ejsmith) [reply src/Exceptionless.Web/ClientApp/src/routes/(app)/project/[projectId]/stacks/[stackId]/+page.svelte]: Yep, this is now extracted into the shared route helper in src/routes/(app)/redirect-to-events.svelte.ts. getEventsNavigationOptionsForFilter documents the stack drilldown behavior, and redirectToEventsWithFilter/navigateToListPage handle the actual navigation URL construction.

src/Exceptionless.Web/ClientApp/src/lib/features/shared/table.svelte.ts
ejsmith: Good call. I cleaned this up in 9bea483a3 so pagination resolves the requested state first, normalizes it once with resolvePaginationChange, and then applies a single setPagination update before updating the cursor/offset/memory params. I also added focused coverage with npm run test:unit -- table.svelte for the page-size reset behavior.

PR #2315, opened 2026-06-20
[codex] Optimize event and stack paging totals

ejsmith:
## Summary

- Adds an `includeTotal` query option to event list endpoints so cursor paging can skip expensive total-hit tracking after the initial page.
- Updates event and stack pages to request totals only when they are useful for stable initial pagination metadata.
- Replaces stack total calculation based on `sum_other_doc_count` with an optional `stack_id` cardinality aggregation, and preserves known table page counts when later offset responses omit totals.
- Updates the Foundatio.Repositories.Elasticsearch package to `8.0.0-beta1.search-after-pit-cursors.25` without using PIT.

## Root Cause

Event paging was asking Elasticsearch for exact totals on every before/after request and still returning `X-Result-Count` on cursor hops. Stack paging was also deriving a stack total from terms aggregation spillover document counts, which is not a stable distinct stack count.

## Validation

- `dotnet restore src/Exceptionless.Web/Exceptionless.Web.csproj`
- `dotnet build src/Exceptionless.Web/Exceptionless.Web.csproj -p:UseSharedCompilation=false -p:BuildInParallel=false --no-restore`
- `dotnet build tests/Exceptionless.Tests/Exceptionless.Tests.csproj -p:UseSharedCompilation=false -p:BuildInParallel=false --no-restore`
- `UPDATE_SNAPSHOTS=true dotnet test tests/Exceptionless.Tests/Exceptionless.Tests.csproj -- --filter-class Exceptionless.Tests.Controllers.OpenApiControllerTests`
- `dotnet test tests/Exceptionless.Tests/Exceptionless.Tests.csproj -- --filter-class Exceptionless.Tests.Controllers.OpenApiControllerTests`
- `npm run check`
- `npm run test:unit -- src/lib/features/shared/table.test.ts`
- `git diff --check`

## Notes

- PIT is intentionally not used because there is not a reliable cleanup path and it is not required for this change.
- Aspire was started locally and reported the API and Svelte app healthy on dynamic ports.

### DIFF
--- a/src/Exceptionless.Core/Exceptionless.Core.csproj
+++ b/src/Exceptionless.Core/Exceptionless.Core.csproj
@@ -35,7 +35,7 @@
     <PackageReference Include="Stripe.net" Version="52.0.0" />
     <PackageReference Include="System.DirectoryServices" Version="10.0.9" />
     <PackageReference Include="UAParser" Version="3.1.47" />
-    <PackageReference Include="Foundatio.Repositories.Elasticsearch" Version="8.0.0-beta1.11" Condition="'$(ReferenceFoundatioRepositoriesSource)' == '' OR '$(ReferenceFoundatioRepositoriesSource)' == 'false'" />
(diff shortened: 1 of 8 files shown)


ejsmith:
Closing in favor of #2316 with corrected branch and PR title naming.

PR #2316, opened 2026-06-20
Optimize event and stack paging totals

ejsmith:
## Summary

- Adds an `include=total` query option to event list endpoints so cursor paging can opt into total metadata only when needed.
- Updates event and stack pages to request totals only when they are useful for stable initial pagination metadata.
- Replaces stack total calculation based on `sum_other_doc_count` with an optional `stack_id` cardinality aggregation, and preserves known table page counts when later offset responses omit totals.
- Updates the Foundatio.Repositories.Elasticsearch package to `8.0.0-beta2` without using PIT.

## Root Cause

Event paging was asking Elasticsearch for exact totals on every before/after request and still returning `X-Result-Count` on cursor hops. Stack paging was also deriving a stack total from terms aggregation spillover document counts, which is not a stable distinct stack count.

## Validation

- `dotnet restore src/Exceptionless.Web/Exceptionless.Web.csproj`
- `dotnet build src/Exceptionless.Web/Exceptionless.Web.csproj -p:UseSharedCompilation=false -p:BuildInParallel=false --no-restore`
- `dotnet build tests/Exceptionless.Tests/Exceptionless.Tests.csproj -p:UseSharedCompilation=false -p:BuildInParallel=false --no-restore`
- `UPDATE_SNAPSHOTS=true dotnet test tests/Exceptionless.Tests/Exceptionless.Tests.csproj -- --filter-class Exceptionless.Tests.Controllers.OpenApiControllerTests`
- `dotnet test tests/Exceptionless.Tests/Exceptionless.Tests.csproj -- --filter-class Exceptionless.Tests.Controllers.OpenApiControllerTests`
- `npm run check`
- `npm run test:unit -- src/lib/features/shared/table.test.ts`
- `git diff --check`
- `UPDATE_SNAPSHOTS=true dotnet test tests/Exceptionless.Tests/Exceptionless.Tests.csproj -- --filter-class Exceptionless.Tests.Controllers.OpenApiControllerTests`
- `dotnet test tests/Exceptionless.Tests/Exceptionless.Tests.csproj -- --filter-class Exceptionless.Tests.Controllers.OpenApiControllerTests`
- `dotnet build src/Exceptionless.Web/Exceptionless.Web.csproj -p:UseSharedCompilation=false -p:BuildInParallel=false --no-restore`
- `npm run check`

## Notes

- PIT is intentionally not used because there is not a reliable cleanup path and it is not required for this change.
- Missing `include` does not include total metadata; clients must pass `include=total` when they need `X-Result-Count`.
- Aspire was started locally and reported the API and Svelte app healthy on dynamic ports.

### DIFF
--- a/docs/serialization-architecture.md
+++ b/docs/serialization-architecture.md
@@ -386,8 +386,8 @@ During deployment where old instances use Newtonsoft and new instances use STJ:
 ### Newtonsoft.Json (Transitive Only — NOT USED)
 
 ```
-Foundatio.Repositories.Elasticsearch v8.0.0-beta1
-  → Foundatio.Repositories v8.0.0-beta1
+Foundatio.Repositories.Elasticsearch v8.0.0-beta2
+  → Foundatio.Repositories v8.0.0-beta2
     → Foundatio.JsonNet v13.0.1
       → Newtonsoft.Json v13.0.4
 
--- a/src/Exceptionless.Core/Exceptionless.Core.csproj
+++ b/src/Exceptionless.Core/Exceptionless.Core.csproj
@@ -35,7 +35,7 @@
     <PackageReference Include="Stripe.net" Version="52.0.0" />
     <PackageReference Include="System.DirectoryServices" Version="10.0.9" />
     <PackageReference Include="UAParser" Version="3.1.47" />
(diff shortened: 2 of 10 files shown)


niemyjski (approved):
This works, we may have to tweak include for CountResult/FindResult as that has total in the response and if we treat this as a field mask and special cases filtering headers it could be inconsistent behavior.

PR #2320, opened 2026-06-22
Add hosted MCP server

ejsmith:
## Summary
- add a hosted MCP server endpoint at `/mcp` inside the existing Exceptionless app so AI tools can query a user's Exceptionless projects directly
- expose MCP read tools for projects, stacks, project-wide events, stack events, event details, event counts, grouped event counts, and supported filter metadata
- support bounded `count_events` grouping for version, tag, stack, user, error type, and other approved dimensions so agents can answer questions like errors by version or whether errors changed since a release window
- add scoped stack write tools to update stack status, mark fixed with a version, snooze stacks, control whether future events are critical, add reference links, and remove reference links
- add OAuth-based authentication for MCP clients, including authorization-code + PKCE, refresh/revoke, protected-resource metadata, DCR, and CIMD client discovery
- refresh system-discovered CIMD client metadata so existing clients can pick up newly supported scopes like `stacks:write`
- store MCP OAuth client applications in Elasticsearch and add management in the new Svelte system settings UI, including the optional `stacks:write` scope
- keep `/mcp` from falling through to the web app shell by returning the correct OAuth challenge or MCP response
- return structured MCP errors for invalid filters, invalid sorts, invalid cursors, invalid write inputs, and missing-scope `forbidden` cases

## Testing
- `dotnet build tests\Exceptionless.Tests\Exceptionless.Tests.csproj -v:minimal -m:1`
- `dotnet test tests\Exceptionless.Tests\Exceptionless.Tests.csproj -- --filter-class Exceptionless.Tests.Controllers.ExceptionlessMcpToolsTests`
- `dotnet test tests\Exceptionless.Tests\Exceptionless.Tests.csproj -- --filter-class Exceptionless.Tests.Controllers.OAuthControllerTests`
- `dotnet test tests\Exceptionless.Tests\Exceptionless.Tests.csproj -- --filter-class Exceptionless.Tests.Controllers.OAuthControllerTests --filter-class Exceptionless.Tests.Controllers.OAuthApplicationControllerTests`
- `dotnet test tests\Exceptionless.Tests\Exceptionless.Tests.csproj -- --filter-class Exceptionless.Tests.Controllers.ControllerManifestTests`
- `dotnet test tests\Exceptionless.Tests\Exceptionless.Tests.csproj -- --filter-class Exceptionless.Tests.Controllers.OAuthApplicationControllerTests --filter-class Exceptionless.Tests.Controllers.OAuthControllerTests --filter-class Exceptionless.Tests.Controllers.OpenApiControllerTests`
- `dotnet test tests\Exceptionless.Tests\Exceptionless.Tests.csproj --no-restore -- --filter-class Exceptionless.Tests.Controllers.OAuthControllerTests`
- `npm run lint` in `src/Exceptionless.Web/ClientApp`
- `npm run check` in `src/Exceptionless.Web/ClientApp`
- `npm run build` in `src/Exceptionless.Web/ClientApp`
- local Aspire smoke test for OAuth + MCP tool calls against `/mcp`
- local Aspire smoke test for unauthenticated/authenticated `GET /mcp` behavior
- dev preview smoke test for `/mcp` OAuth protected-resource discovery metadata and MCP read/write tools
- `git diff --check`
- `rg "<<<<<<<|=======|>>>>>>>"`
- GitHub PR checks are passing

## Notes
- OAuth is included to let external MCP clients authenticate to the hosted Exceptionless MCP server; it is not the primary feature by itself.
- The default OAuth authorization flow remains read-oriented; MCP clients must explicitly request `stacks:write` before stack mutation tools can run.
- CIMD clients discovered before new scopes were added are refreshed from their metadata document during authorization; manually managed and DCR clients keep their configured scopes.
- Breaking changes: none.


src/Exceptionless.Core/Models/OAuthApplication.cs
```diff
@@ -0,0 +1,74 @@
...
+
+    [ObjectId]
+    public string? UpdatedByUserId { get; set; }
+
+    public DateTime CreatedUtc { get; set; }
+    public DateTime UpdatedUtc { get; set; }
+
+    public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
+    {
+        foreach (string redirectUri in RedirectUris.Where(String.IsNullOrWhiteSpace))
```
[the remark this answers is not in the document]
ejsmith: Addressed in 47a079937 and 81184208a: the unused validation loop variables now use discards, and the validation loops now filter directly in the sequence.

tests/Exceptionless.Tests/Controllers/OAuthControllerTests.cs
```diff
@@ -0,0 +1,421 @@
...
+        var token = await IssueTokenAsync();
+        Assert.NotNull(token.RefreshToken);
+        using var client = CreateHttpClient();
+
+        var refreshResponse = await client.PostAsync("oauth/token", new FormUrlEncodedContent(new Dictionary<string, string?>
+        {
+            ["grant_type"] = OAuthGrantTypes.RefreshToken,
+            ["client_id"] = ClientId,
+            ["refresh_token"] = token.RefreshToken
+        }), TestContext.Current.CancellationToken);
```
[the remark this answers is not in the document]
ejsmith: Addressed in 47a079937: the refresh-token FormUrlEncodedContent instances are now using var locals before PostAsync.

tests/Exceptionless.Tests/Controllers/OAuthControllerTests.cs
```diff
@@ -0,0 +1,421 @@
...
+        Assert.NotNull(refreshedToken);
+        Assert.NotEqual(token.AccessToken, refreshedToken.AccessToken);
+        Assert.NotEqual(token.RefreshToken, refreshedToken.RefreshToken);
+
+        var reusedRefreshResponse = await client.PostAsync("oauth/token", new FormUrlEncodedContent(new Dictionary<string, string?>
+        {
+            ["grant_type"] = OAuthGrantTypes.RefreshToken,
+            ["client_id"] = ClientId,
+            ["refresh_token"] = token.RefreshToken
+        }), TestContext.Current.CancellationToken);
```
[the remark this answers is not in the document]
ejsmith: Addressed in 47a079937: the reused refresh-token FormUrlEncodedContent is now a using var local before PostAsync.

tests/Exceptionless.Tests/Controllers/OAuthControllerTests.cs
```diff
@@ -0,0 +1,421 @@
...
+    [Fact]
+    public async Task RevokeAsync_DisablesOAuthToken()
+    {
+        var token = await IssueTokenAsync();
+        using var client = CreateHttpClient();
+
+        var response = await client.PostAsync("oauth/revoke", new FormUrlEncodedContent(new Dictionary<string, string?>
+        {
+            ["token"] = token.AccessToken
+        }), TestContext.Current.CancellationToken);
```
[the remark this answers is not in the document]
ejsmith: Addressed in 47a079937: the revoke FormUrlEncodedContent is now a using var local before PostAsync.

src/Exceptionless.Core/Models/OAuthApplication.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        foreach (string scope in Scopes.Where(String.IsNullOrWhiteSpace))
+            yield return new ValidationResult("Scope cannot be empty.", [nameof(Scopes)]);
+
+        foreach (string scope in Scopes.Where(s => !String.IsNullOrWhiteSpace(s)))
+        {
+            if (!AllScopes.Contains(scope, StringComparer.Ordinal))
+                yield return new ValidationResult($"'{scope}' is not a supported OAuth scope.", [nameof(Scopes)]);
+        }
```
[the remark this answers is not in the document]
ejsmith: Addressed in 81184208a: the scope validation loop now filters unsupported scopes directly in the sequence.

src/Exceptionless.Core/Services/OAuthService.cs
```diff
@@ -0,0 +1,579 @@
...
+                continue;
+
+            if (!registeredUri.IsDefaultPort && registeredUri.Port != requestedUri.Port)
+                continue;
+
+            if (String.Equals(registeredUri.Host, requestedUri.Host, StringComparison.OrdinalIgnoreCase)
+                && String.Equals(registeredUri.AbsolutePath, requestedUri.AbsolutePath, StringComparison.Ordinal)
+                && String.Equals(registeredUri.Query, requestedUri.Query, StringComparison.Ordinal))
+                return true;
+        }
```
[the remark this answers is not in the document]
ejsmith: Addressed in 47a079937: IsRedirectUriAllowed now parses and filters allowed redirect URIs in the sequence before the final equality check.

tests/Exceptionless.Tests/Controllers/ExceptionlessMcpToolsTests.cs
```diff
@@ -0,0 +1,499 @@
...
+        var tools = await CreateToolsAsync(AuthorizationRoles.McpRead, AuthorizationRoles.EventsRead);
+
+        var firstPage = await tools.GetStackEventsAsync(stacks[0].Id, limit: 1);
+
+        Assert.True(firstPage.Ok);
+        Assert.Null(firstPage.Error);
+        Assert.True(firstPage.Pagination?.HasMore);
+        Assert.NotNull(firstPage.Pagination?.After);
+
+        var secondPage = await tools.GetStackEventsAsync(stacks[0].Id, limit: 1, after: firstPage.Pagination.After);
```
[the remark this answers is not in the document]
ejsmith: Addressed in 47a079937: the pagination value is now captured and asserted non-null before using the cursor.

src/Exceptionless.Core/Services/OAuthService.cs
```diff
@@ -0,0 +1,618 @@
...
+            .Where(uri => uri is not null)
+            .Select(uri => uri!)
+            .Where(uri => IsLoopbackHttpRedirectUri(uri) && IsLoopbackHttpRedirectUri(requestedUri))
+            .Where(uri => uri.IsDefaultPort || uri.Port == requestedUri.Port))
+        {
+            if (String.Equals(registeredUri.Host, requestedUri.Host, StringComparison.OrdinalIgnoreCase)
+                && String.Equals(registeredUri.AbsolutePath, requestedUri.AbsolutePath, StringComparison.Ordinal)
+                && String.Equals(registeredUri.Query, requestedUri.Query, StringComparison.Ordinal))
+                return true;
+        }
```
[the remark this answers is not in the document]
ejsmith: Declined: keeping the imperative loop here. The continue-based validation is easier to read than folding every predicate into the foreach source.

src/Exceptionless.Web/Mcp/ExceptionlessMcpTools.cs
```diff
@@ -0,0 +1,1856 @@
...
+        }
+
+        foreach (string field in sort.Split([',', ' '], StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries).Select(term => term.TrimStart('+', '-')))
+        {
+            if (field.Length == 0 || !allowedSortFields.Contains(field))
+            {
+                error = $"Unknown sort field '{field}'. Allowed sort fields: {String.Join(", ", allowedSortFields.Order(StringComparer.OrdinalIgnoreCase))}.";
+                return false;
+            }
+        }
```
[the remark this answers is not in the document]
ejsmith: Declined: keeping the local field assignment in the loop. It reads better than adding a LINQ pipeline solely to satisfy a style suggestion.

src/Exceptionless.Web/Mcp/ExceptionlessMcpTools.cs
```diff
@@ -0,0 +1,1856 @@
...
+
+        foreach (Match match in FilterFieldRegex.Matches(filter))
+        {
+            string field = match.Groups["field"].Value;
+            if (field.StartsWith("data.", StringComparison.OrdinalIgnoreCase) || field.StartsWith("idx.", StringComparison.OrdinalIgnoreCase))
+                continue;
+
+            if (!allowedFilterFields.Contains(field))
+                return field;
+        }
```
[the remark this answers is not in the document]
ejsmith: Declined: keeping the Match loop and local field extraction. This is clearer for regex group handling than projecting the MatchCollection in the foreach header.

src/Exceptionless.Core/Services/OAuthClientMetadataService.cs
```diff
@@ -0,0 +1,167 @@
...
+        }
+        catch (OperationCanceledException)
+        {
+            return null;
+        }
+        catch (Exception ex)
+        {
+            logger.LogWarning(ex, "Unable to fetch OAuth client metadata document for {ClientId}.", clientId);
+            return null;
+        }
```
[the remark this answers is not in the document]
ejsmith: Addressed in b50ee7eb3: replaced the broad metadata catch with specific expected exception catches and shared logging.

src/Exceptionless.Web/Mcp/ExceptionlessMcpTools.cs
```diff
@@ -0,0 +1,1856 @@
...
+            return McpResponse<McpListData<McpProjectResult>>.Failed(McpErrors.NotAccessible("Unable to list projects. No accessible organizations were found."));
+        }
+        catch (Exception ex) when (!String.IsNullOrWhiteSpace(filter))
+        {
+            return McpResponse<McpListData<McpProjectResult>>.Failed(McpErrors.InvalidFilter($"Invalid filter: {ex.Message}"));
+        }
+        catch (Exception)
+        {
+            return McpResponse<McpListData<McpProjectResult>>.Failed(McpErrors.QueryFailed("Unable to list projects. Check the filter, sort, and limit values."));
+        }
```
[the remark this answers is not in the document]
ejsmith: Addressed in b50ee7eb3: narrowed the MCP list_projects fallback catch to expected tool/query errors.

src/Exceptionless.Web/Mcp/ExceptionlessMcpTools.cs
```diff
@@ -0,0 +1,1856 @@
...
+            return ToListResponse(results, ToStackResult, validation.Warning, validation.Limit);
+        }
+        catch (Exception ex) when (IsLookupError(ex))
+        {
+            return McpResponse<McpListData<McpStackResult>>.Failed(ToLookupError("Project", projectId, ex));
+        }
+        catch (Exception)
+        {
+            return McpResponse<McpListData<McpStackResult>>.Failed(McpErrors.QueryFailed("Unable to search stacks. Check the project id, filter, sort, and limit values."));
+        }
```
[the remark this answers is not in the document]
ejsmith: Addressed in b50ee7eb3: narrowed the search_stacks fallback catch to expected tool/query errors.

src/Exceptionless.Web/Mcp/ExceptionlessMcpTools.cs
```diff
@@ -0,0 +1,1856 @@
...
+            return ToListResponse(results, ev => ToEventResult(ev), validation.Warning, validation.Limit);
+        }
+        catch (Exception ex) when (IsLookupError(ex))
+        {
+            return McpResponse<McpListData<McpEventResult>>.Failed(ToLookupError("Stack", stackId, ex));
+        }
+        catch (Exception)
+        {
+            return McpResponse<McpListData<McpEventResult>>.Failed(McpErrors.QueryFailed("Unable to list stack events. Check the stack id, filter, sort, and limit values."));
+        }
```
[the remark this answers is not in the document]
ejsmith: Addressed in b50ee7eb3: narrowed the get_stack_events fallback catch to expected tool/query errors.

src/Exceptionless.Web/Mcp/ExceptionlessMcpTools.cs
```diff
@@ -0,0 +1,1856 @@
...
+            return ToListResponse(results, ev => ToEventResult(ev), validation.Warning, validation.Limit);
+        }
+        catch (Exception ex) when (IsLookupError(ex))
+        {
+            return McpResponse<McpListData<McpEventResult>>.Failed(ToLookupError("Project", projectId, ex));
+        }
+        catch (Exception)
+        {
+            return McpResponse<McpListData<McpEventResult>>.Failed(McpErrors.QueryFailed("Unable to search events. Check the project id, filter, sort, limit, and time range values."));
+        }
```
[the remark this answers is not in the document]
ejsmith: Addressed in b50ee7eb3: narrowed the search_events fallback catch to expected tool/query errors.

src/Exceptionless.Web/Mcp/ExceptionlessMcpTools.cs
```diff
@@ -0,0 +1,1856 @@
...
+                groupLimitWarning);
+        }
+        catch (Exception ex) when (IsLookupError(ex))
+        {
+            return McpResponse<McpEventCountResult>.Failed(ToLookupError("Project", projectId, ex));
+        }
+        catch (Exception)
+        {
+            return McpResponse<McpEventCountResult>.Failed(McpErrors.QueryFailed("Unable to count events. Check the project id, filter, time range, and interval values."));
+        }
```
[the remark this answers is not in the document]
ejsmith: Addressed in b50ee7eb3: narrowed the count_events fallback catch to expected tool/query errors.

src/Exceptionless.Web/Mcp/ExceptionlessMcpTools.cs
```diff
@@ -0,0 +1,1856 @@
...
+                $"Stack {stack.Id} status is {stack.Status.ToString().ToLowerInvariant()}."));
+        }
+        catch (Exception ex) when (IsLookupError(ex))
+        {
+            return McpResponse<McpStackUpdateResult>.Failed(ToLookupError("Stack", stackId, ex));
+        }
+        catch (Exception)
+        {
+            return McpResponse<McpStackUpdateResult>.Failed(McpErrors.QueryFailed("Unable to update stack status. Check the stack id, status, and fixed version."));
+        }
```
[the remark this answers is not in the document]
ejsmith: Addressed in b50ee7eb3: narrowed the update_stack_status fallback catch to expected tool/query errors.

src/Exceptionless.Core/Configuration/AppOptions.cs
```diff
@@ -131,6 +132,7 @@ public static AppOptions ReadFromConfiguration(IConfiguration config)
         options.SlackOptions = SlackOptions.ReadFromConfiguration(config);
         options.StripeOptions = StripeOptions.ReadFromConfiguration(config);
         options.AuthOptions = AuthOptions.ReadFromConfiguration(config);
+        options.OAuthOptions = OAuthOptions.ReadFromConfiguration(config);
```
niemyjski: AuthOptions has OAuth parsing in it now, so this is kind of confusing.

src/Exceptionless.Core/Models/OAuthApplication.cs
```diff
@@ -0,0 +1,86 @@
+using System.ComponentModel.DataAnnotations;
+using Exceptionless.Core.Attributes;
+using Exceptionless.Core.Authorization;
+using Foundatio.Repositories.Models;
+
+namespace Exceptionless.Core.Models;
+
+public class OAuthApplication : IIdentity, IHaveDates, IValidatableObject
+{
+    public const string SystemUserId = "000000000000000000000000";
```
niemyjski: Is this supposed to be like an ObjectId?

src/Exceptionless.Core/Models/OAuthApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public static readonly IReadOnlyCollection<string> AllScopes =
+    [
+        AuthorizationRoles.McpRead,
+        AuthorizationRoles.ProjectsRead,
+        AuthorizationRoles.StacksRead,
+        AuthorizationRoles.StacksWrite,
+        AuthorizationRoles.EventsRead,
+        AuthorizationRoles.OfflineAccess
```
niemyjski, line 73: What is OfflineAccess for a connected online service?

src/Exceptionless.Core/Configuration/AppOptions.cs
ejsmith: Fixed in 46588885e by renaming the new hosted OAuth server settings to OAuthServerOptions and moving their config section to OAuthServer: so it no longer overlaps conceptually with AuthOptions parsing ConnectionStrings:OAuth for external login providers.

src/Exceptionless.Core/Models/OAuthApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+    public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
+    {
+        foreach (string _ in RedirectUris.Where(String.IsNullOrWhiteSpace))
+            yield return new ValidationResult("Redirect URI cannot be empty.", [nameof(RedirectUris)]);
+
+        foreach (string redirectUri in RedirectUris.Where(uri => !String.IsNullOrWhiteSpace(uri)))
+        {
+            if (!IsValidRedirectUri(redirectUri))
+                yield return new ValidationResult($"'{redirectUri}' must be an absolute HTTPS URI or loopback HTTP URI without a fragment.", [nameof(RedirectUris)]);
+        }
```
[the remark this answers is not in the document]
ejsmith: Declined: keeping the imperative validation body here. It is clearer than moving every validation predicate into chained Where calls.

src/Exceptionless.Core/Models/OAuthApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        foreach (string _ in Scopes.Where(String.IsNullOrWhiteSpace))
+            yield return new ValidationResult("Scope cannot be empty.", [nameof(Scopes)]);
+
+        foreach (string scope in Scopes.Where(s => !String.IsNullOrWhiteSpace(s)))
+        {
+            if (!AllScopes.Contains(scope, StringComparer.Ordinal))
+                yield return new ValidationResult($"'{scope}' is not a supported OAuth scope.", [nameof(Scopes)]);
+        }
```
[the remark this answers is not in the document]
ejsmith: Declined: keeping the imperative validation body here. It is clearer than moving every validation predicate into chained Where calls.

src/Exceptionless.Core/Services/OAuthService.cs
```diff
@@ -0,0 +1,622 @@
...
+                continue;
+
+            if (!registeredUri.IsDefaultPort && registeredUri.Port != requestedUri.Port)
+                continue;
+
+            if (String.Equals(registeredUri.Host, requestedUri.Host, StringComparison.OrdinalIgnoreCase)
+                && String.Equals(registeredUri.AbsolutePath, requestedUri.AbsolutePath, StringComparison.Ordinal)
+                && String.Equals(registeredUri.Query, requestedUri.Query, StringComparison.Ordinal))
+                return true;
+        }
```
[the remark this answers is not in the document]
ejsmith: Declined: keeping the continue-based redirect URI validation. The imperative form is easier to scan than folding all checks into the foreach source.

src/Exceptionless.Web/Mcp/ExceptionlessMcpTools.cs
```diff
@@ -0,0 +1,1862 @@
...
+
+        foreach (string term in sort.Split([',', ' '], StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries))
+        {
+            string field = term.TrimStart('+', '-');
+            if (field.Length == 0 || !allowedSortFields.Contains(field))
+            {
+                error = $"Unknown sort field '{field}'. Allowed sort fields: {String.Join(", ", allowedSortFields.Order(StringComparer.OrdinalIgnoreCase))}.";
+                return false;
+            }
+        }
```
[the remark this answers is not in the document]
ejsmith: Declined: keeping the local field assignment. It is clearer than projecting in the foreach header just to satisfy a style-only suggestion.

src/Exceptionless.Core/Services/OAuthClientMetadataService.cs
```diff
@@ -0,0 +1,191 @@
...
+        return memoryStream;
+    }
+
+    private static async Task<bool> IsPublicHostAsync(Uri uri, CancellationToken cancellationToken)
+    {
+        IPAddress[] addresses;
+        if (IPAddress.TryParse(uri.Host, out var address))
+            addresses = [address];
+        else
+            addresses = await Dns.GetHostAddressesAsync(uri.Host, cancellationToken);
```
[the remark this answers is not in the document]
ejsmith: Declined: keeping the explicit if/else here. It is more readable than a ternary for this parsing path.

src/Exceptionless.Core/Services/OAuthClientMetadataService.cs
```diff
@@ -0,0 +1,191 @@
...
+        try
+        {
+            using var cts = new CancellationTokenSource(options.ClientMetadataDocumentRequestTimeout);
+            if (!await IsPublicHostAsync(uri, cts.Token))
+                return null;
+
+            using var request = new HttpRequestMessage(HttpMethod.Get, uri);
+            request.Headers.Accept.ParseAdd("application/json");
+
+            using var response = await httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cts.Token);
```
niemyjski, line 44: do we need to do any backoffs here or possibly disallow retries after consecutive failures?

src/Exceptionless.Core/Models/OAuthApplication.cs
ejsmith: Yes, it needs to satisfy the ObjectId-shaped audit fields. I changed it in 55059b196 from the all-zero sentinel to the existing system ObjectId pattern ending in 1, which makes the intent less ambiguous.

src/Exceptionless.Core/Models/OAuthApplication.cs
ejsmith: offline_access is only used to decide whether we issue a refresh token. It matters for CLI/desktop MCP clients like Claude Code or Codex so they do not have to send users through the browser auth flow every hour. Online clients can omit it and will just receive an access token.

src/Exceptionless.Core/Services/OAuthClientMetadataService.cs
```diff
@@ -0,0 +1,191 @@
...
+        catch (CryptographicException ex)
+        {
+            LogUnableToFetchMetadata(ex, clientId);
+            return null;
+        }
+        catch (InvalidOperationException ex)
+        {
+            LogUnableToFetchMetadata(ex, clientId);
+            return null;
+        }
```
niemyjski: Could we combine all of these?

src/Exceptionless.Core/Services/OAuthClientMetadataService.cs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    private void LogUnableToFetchMetadata(Exception ex, string clientId)
+    {
+        logger.LogWarning(ex, "Unable to fetch OAuth client metadata document for {ClientId}.", clientId);
+    }
+    public static bool TryCreateClientMetadataDocumentUri(string clientId, out Uri uri)
```
niemyjski, line 87:
```suggestion
    }

    public static bool TryCreateClientMetadataDocumentUri(string clientId, out Uri uri)
```

src/Exceptionless.Core/Services/OAuthService.cs
```diff
@@ -0,0 +1,622 @@
...
+            .Select(uri => uri.Trim())
+            .Distinct(StringComparer.Ordinal)
+            .ToArray();
+
+        if (redirectUris.Length == 0 || redirectUris.Length > 20 || redirectUris.Any(uri => !OAuthApplication.IsValidRedirectUri(uri)))
+            return OAuthClientRegistrationResult.Invalid("invalid_redirect_uri", "Redirect URIs must be absolute HTTPS URIs or loopback HTTP URIs without fragments.");
+
+        var scopes = NormalizeScopes(request.Scope);
+        if (scopes.Count == 0)
+            scopes = SupportedScopes;
```
niemyjski: Defaulting an untrusted public client to every supported scope is too permissive. An anonymous DCR request with no `scope` becomes allowed for `stacks:write` and `offline_access`, and the metadata-doc path below does the same at line 189. That makes the broadest grant the default instead of an explicit client/admin declaration. Please default dynamic clients to the minimal read-only scope set, or require explicit/admin-managed approval before write/offline scopes are allowed.

src/Exceptionless.Core/Services/OAuthService.cs
```diff
@@ -0,0 +1,622 @@
...
+        if (client is null)
+            return OAuthValidationResult.Invalid("invalid_client", "Unknown OAuth client.");
+
+        if (!String.Equals(request.ResponseType, "code", StringComparison.Ordinal))
+            return OAuthValidationResult.Invalid("unsupported_response_type", "Only the code response type is supported.");
+
+        if (!IsRedirectUriAllowed(request.RedirectUri, client.RedirectUris))
+            return OAuthValidationResult.Invalid("invalid_request", "Invalid redirect_uri.");
+
+        if (String.IsNullOrWhiteSpace(request.CodeChallenge) || !String.Equals(request.CodeChallengeMethod, CodeChallengeMethod, StringComparison.Ordinal))
```
niemyjski: This only checks presence and `S256`; it does not validate the PKCE challenge/verifier shape. RFC 7636 requires a high-entropy verifier of 43-128 unreserved characters, and an S256 challenge should be a base64url SHA256 value. The current tests use `valid-test-code-verifier` (24 chars), which demonstrates that a weak verifier is accepted. Please validate `code_challenge` syntax at authorization and `code_verifier` length/charset before hashing at token exchange. Reference: https://datatracker.ietf.org/doc/html/rfc7636

src/Exceptionless.Core/Services/OAuthService.cs
```diff
@@ -0,0 +1,622 @@
...
+
+    public async Task<OAuthTokenIssueResult> RefreshAsync(OAuthTokenRequest request)
+    {
+        if (String.IsNullOrWhiteSpace(request.RefreshToken) || String.IsNullOrWhiteSpace(request.ClientId))
+            return OAuthTokenIssueResult.Invalid("invalid_request", "Missing refresh_token or client_id.");
+
+        if (await GetClientAsync(request.ClientId) is null)
+            return OAuthTokenIssueResult.Invalid("invalid_client", "Unknown OAuth client.");
+
+        var results = await tokenRepository.GetByRefreshTokenAsync(request.RefreshToken, o => o.ImmediateConsistency());
```
niemyjski, line 373: Refresh-token rotation needs to be single-use under concurrency. This read/save/create sequence is not atomic, so two concurrent refresh requests can both read the same still-enabled token and each mint a new token before either save wins. Please guard this with a per-refresh-token lock or optimistic concurrency check/retry, and add a concurrent refresh test. Ideally refresh-token reuse should revoke the token family.

src/Exceptionless.Web/ClientApp/src/routes/(auth)/oauth/authorize/+page.svelte
```diff
@@ -0,0 +1,188 @@
...
+                    <Muted>Resource</Muted>
+                    <p class="break-all font-mono text-xs">{resource}</p>
+                </div>
+                <div>
+                    <Muted>Scopes</Muted>
+                    <div class="mt-2 flex flex-wrap gap-1.5">
+                        {#each requestedScopes as scope (scope)}
+                            <Badge variant="secondary">{scope}</Badge>
+                        {:else}
+                            <Badge variant="outline">Default MCP read access</Badge>
```
niemyjski: This consent text understates the grant. When `scope` is omitted, the backend defaults to `mcp:read projects:read stacks:read events:read`, but the page tells the user only `Default MCP read access`. OAuth consent must show the exact permissions that will be issued into the authorization code/token. Please fetch validated authorization details from the API or mirror the backend defaults exactly here.

src/Exceptionless.Core/Services/OAuthClientMetadataService.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        string cacheKey = GetCacheKey(clientId);
+        var cached = await cacheClient.GetAsync<OAuthClientMetadataDocument>(cacheKey);
+        if (cached.HasValue)
+            return cached.Value;
+
+        try
+        {
+            using var cts = new CancellationTokenSource(options.ClientMetadataDocumentRequestTimeout);
+            if (!await IsPublicHostAsync(uri, cts.Token))
```
niemyjski: This DNS preflight is still a TOCTOU SSRF control: `Dns.GetHostAddressesAsync` validates one resolution, then `HttpClient` resolves/connects separately. A client metadata URL controlled by an attacker can DNS-rebind after validation and target internal/metadata addresses. Because this fetches attacker-controlled URLs server-side, please enforce the public-address check at the actual connection/egress layer or restrict metadata fetching to an allowlist. Also extend `IsPublicAddress` for non-global ranges such as 100.64/10 and 198.18/15. Reference: https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html

src/Exceptionless.Web/Controllers/OAuthController.cs
```diff
@@ -0,0 +1,327 @@
...
+    [AllowAnonymous]
+    public async Task<IActionResult> RevokeAsync([FromForm] OAuthRevokeForm form)
+    {
+        await oauthService.RevokeAsync(form.Token);
+        return Ok();
+    }
+
+    private string GetOrigin()
+    {
+        return $"{Request.Scheme}://{Request.Host}";
```
niemyjski: OAuth discovery metadata should use a configured canonical external origin, not `Request.Host`. Clients trust these metadata URLs (`issuer`, authorization/token endpoints, and resource identifiers), and deriving them from the Host header can produce spoofed metadata or token/resource mismatches when a proxy passes an unexpected host. Please build this from `AppOptions.BaseURL` or another canonical public origin.

src/Exceptionless.Web/Controllers/OAuthController.cs
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    [HttpPost(API_PREFIX + "/oauth/authorize")]
+    [Authorize(Policy = AuthorizationRoles.UserPolicy)]
+    public Task<IActionResult> CompleteAuthorizeAsync([FromBody] OAuthAuthorizeForm form)
+    {
+        return CompleteAuthorizationAsync(form.ToRequest(), jsonResponse: true);
+    }
+
+    [HttpPost(API_PREFIX + "/oauth/register")]
```
niemyjski, line 76: This is an anonymous endpoint that creates durable `OAuthApplication` documents. The only protection I found is the broad API throttle, which is high and fail-open if cache increment fails. Since DCR can be abused for persistent storage/cardinality growth, please add endpoint-specific abuse controls (tighter rate limit, cap/TTL cleanup, feature flag) or require authenticated/admin registration for persisted clients.

src/Exceptionless.Web/Mcp/ExceptionlessMcpTools.cs
```diff
@@ -0,0 +1,1862 @@
...
+using Foundatio.Repositories.Extensions;
+using Foundatio.Repositories.Models;
+using Foundatio.Serializer;
+using Microsoft.Extensions.Logging;
+using ModelContextProtocol.Server;
+
+namespace Exceptionless.Web.Mcp;
+
+[McpServerToolType]
+public sealed class ExceptionlessMcpTools
```
niemyjski, line 28: This new tool surface is 1,862 lines and mixes tool orchestration, authorization, query validation, aggregation building, result projection, error contracts, and DTO records in one file. That is too much for a new security-sensitive API boundary. Please split it before merge, for example into read tools, stack-write tools, query/time validation, result mappers, and MCP error/DTO contracts. Smaller boundaries will make scope/access checks auditable.

niemyjski (changes requested):
Requesting changes. I focused on the OAuth/MCP trust boundary and found several issues that should be fixed before this merges: PKCE verifier/challenge validation is below RFC 7636 requirements, the consent screen can underreport the scopes that are actually granted, dynamic clients default to every scope including write/offline access, client metadata fetching still has an SSRF/DNS-rebinding gap, refresh-token rotation is not concurrency-safe, OAuth metadata derives its origin from Request.Host, and anonymous DCR needs stronger abuse controls. I also called out the new MCP tools file because it is already too large and mixed for a security-sensitive API surface.

Local validation I ran:
- `dotnet test -- --filter-class Exceptionless.Tests.Controllers.OAuthControllerTests --parallel none` - 30 passed
- `dotnet test -- --filter-class Exceptionless.Tests.Controllers.ExceptionlessMcpToolsTests --parallel none` - 67 passed
- `git diff --check origin/main...HEAD` - clean

GitHub checks at review time: client tests, JavaScript analysis, CLA, version, and preview jobs were passing; API/docker/C# analysis were still pending.

src/Exceptionless.Core/Services/OAuthClientMetadataService.cs
```diff
@@ -0,0 +1,191 @@
...
+public interface IOAuthClientMetadataService
+{
+    Task<OAuthClientMetadataDocument?> GetClientMetadataAsync(string clientId);
+}
+
+public sealed class OAuthClientMetadataService(HttpClient httpClient, OAuthServerOptions options, ICacheClient cacheClient, ILogger<OAuthClientMetadataService> logger) : IOAuthClientMetadataService
+{
+    private const string CachePrefix = "oauth:cimd:";
+
+    private static readonly JsonSerializerOptions JsonSerializerOptions = new(JsonSerializerDefaults.Web);
```
niemyjski: We use web defaults here but do we need to be, we are already decorating the model with JsonPropertyName attributes.

src/Exceptionless.Core/Services/OAuthService.cs
```diff
@@ -0,0 +1,622 @@
...
+using Exceptionless.Core.Repositories;
+using Foundatio.Caching;
+using Foundatio.Repositories;
+using Foundatio.Repositories.Utility;
+
+namespace Exceptionless.Core.Services;
+
+public class OAuthService(OAuthServerOptions options, ICacheClient cacheClient, IOAuthApplicationRepository oauthApplicationRepository, IOAuthClientMetadataService oauthClientMetadataService, ITokenRepository tokenRepository, TimeProvider timeProvider)
+{
+    public const string CodeChallengeMethod = "S256";
```
niemyjski, line 19: should this be a const? where does this number come from?

src/Exceptionless.Core/Services/OAuthService.cs
```diff
@@ -0,0 +1,622 @@
...
+    private const string AuthorizationCodeCachePrefix = "oauth:code:";
+    private const string ClientMetadataNotes = "Discovered from OAuth client metadata document.";
+    private const string DynamicClientRegistrationNotes = "Registered through OAuth dynamic client registration.";
+
+    public bool ClientIdMetadataDocumentSupported => options.EnableClientIdMetadataDocuments;
+
+    public async Task<OAuthClientRegistrationResult> RegisterClientAsync(OAuthClientRegistrationRequest request)
+    {
+        string tokenEndpointAuthMethod = String.IsNullOrWhiteSpace(request.TokenEndpointAuthMethod) ? "none" : request.TokenEndpointAuthMethod.Trim();
+        if (!String.Equals(tokenEndpointAuthMethod, "none", StringComparison.Ordinal))
```
niemyjski, line 54: Can we use minivalidator for this and flow this in from the data annotations?

src/Exceptionless.Core/Services/OAuthService.cs
```diff
@@ -0,0 +1,622 @@
...
+            scopes = SupportedScopes;
+
+        if (scopes.Any(s => !SupportedScopes.Contains(s, StringComparer.Ordinal)))
+            return OAuthClientRegistrationResult.Invalid("invalid_client_metadata", "One or more scopes are not supported.");
+
+        var utcNow = timeProvider.GetUtcNow().UtcDateTime;
+        string clientId = await CreateUniqueClientIdAsync();
+        var application = new OAuthApplication
+        {
+            Id = ObjectId.GenerateNewId().ToString(),
```
niemyjski, line 89: we can get rid of this, this is created for us, smae with created and updated?

src/Exceptionless.Core/Services/OAuthService.cs
```diff
@@ -0,0 +1,622 @@
...
+            ClientId = clientId,
+            Name = NormalizeClientName(request.ClientName, clientId),
+            RedirectUris = redirectUris,
+            Scopes = scopes.ToArray(),
+            Notes = DynamicClientRegistrationNotes,
+            CreatedByUserId = OAuthApplication.SystemUserId,
+            UpdatedByUserId = OAuthApplication.SystemUserId,
```
niemyjski, line 96: Shouldn't we be populating this from http context?

src/Exceptionless.Core/Services/OAuthService.cs
```diff
@@ -0,0 +1,622 @@
...
+    {
+        if (String.IsNullOrWhiteSpace(request.ClientId))
+            return OAuthValidationResult.Invalid("invalid_request", "Missing client_id.");
+
+        var client = await GetClientAsync(request.ClientId, allowClientMetadataDocument: true);
+        if (client is null)
+            return OAuthValidationResult.Invalid("invalid_client", "Unknown OAuth client.");
+
+        if (!String.Equals(request.ResponseType, "code", StringComparison.Ordinal))
+            return OAuthValidationResult.Invalid("unsupported_response_type", "Only the code response type is supported.");
```
niemyjski, line 288: is all this validation duplicated, looks really familiar to the last comment I made on validation driven by data annotations.

src/Exceptionless.Core/Services/OAuthService.cs
```diff
@@ -0,0 +1,622 @@
...
+        token.IsDisabled = true;
+        token.Refresh = null;
+        await tokenRepository.SaveAsync(token, o => o.ImmediateConsistency());
+        return true;
+    }
+
+    private async Task<OAuthTokenResponse> CreateTokenAsync(string userId, string clientId, string resource, IReadOnlyCollection<string> scopes)
+    {
+        var utcNow = timeProvider.GetUtcNow().UtcDateTime;
+        var accessToken = StringExtensions.GetNewToken();
```
niemyjski, line 412: what if this token is already used?

src/Exceptionless.Core/Bootstrapper.cs
```diff
@@ -182,6 +183,9 @@ public static void RegisterServices(IServiceCollection services, AppOptions appO
         services.AddSingleton<NotificationService>();
         services.AddSingleton<OrganizationService>();
         services.AddStartupAction<OrganizationService>();
+        services.AddHttpClient<IOAuthClientMetadataService, OAuthClientMetadataService>(client => client.Timeout = appOptions.OAuthServerOptions.ClientMetadataDocumentRequestTimeout)
```
niemyjski: I could be mistaken but feels like we are configuring the default timeout here as well as when we actually make the http call with the same timeout?

src/Exceptionless.Web/ClientApp/src/lib/features/admin/api.svelte.ts
```diff
@@ -77,6 +104,25 @@ export function getMigrationsQuery() {
     }));
 }
 
+export function getOAuthApplicationsQuery() {
+    return createQuery<OAuthApplication[], ProblemDetails>(() => ({
+        queryFn: async ({ signal }: { signal: AbortSignal }) => {
+            const client = useFetchClient();
+            const response = await client.getJSON<OAuthApplication[]>('admin/oauth-applications', {
```
niemyjski, line 111: shouldn't any customer be able to define an oauth application? If so, admin doesn't feel like the right spot for this.

src/Exceptionless.Web/ClientApp/src/lib/features/admin/models.ts
```diff
@@ -93,6 +93,29 @@ export type MigrationState = {
 
 export type MigrationStatus = 'Completed' | 'Failed' | 'Pending' | 'Running';
 
+export type OAuthApplication = {
```
niemyjski, line 96: these feel like they should be generated via openapi

src/Exceptionless.Web/ClientApp/src/lib/features/admin/schemas.ts
```diff
@@ -7,3 +7,33 @@ export const RunMaintenanceJobSchema = object({
     utcStart: date().optional()
 });
 export type RunMaintenanceJobFormData = Infer<typeof RunMaintenanceJobSchema>;
+
+export const OAuthApplicationSchema = object({
```
niemyjski, line 11: can this be generated via openapi?

src/Exceptionless.Web/Controllers/OAuthController.cs
```diff
@@ -0,0 +1,327 @@
...
+using Exceptionless.Core.Extensions;
+using Exceptionless.Core.Services;
+using Exceptionless.Web.Extensions;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using System.Text.Json.Serialization;
+
+namespace Exceptionless.Web.Controllers;
+
+[Route("")]
```
niemyjski, line 14: feels like this should be versioned + oauth and then the routes below are simplier too.

src/Exceptionless.Web/Controllers/OAuthController.cs
```diff
@@ -0,0 +1,327 @@
...
+            CodeChallenge = CodeChallenge,
+            CodeChallengeMethod = CodeChallengeMethod,
+            Resource = Resource
+        };
+    }
+}
+
+public sealed record OAuthAuthorizeResponse
+{
+
```
niemyjski:
```suggestion
```

src/Exceptionless.Web/Controllers/OAuthController.cs
```diff
@@ -0,0 +1,327 @@
...
+}
+
+public sealed record OAuthErrorResponse
+{
+    [JsonPropertyName("error")]
+    public required string Error { get; init; }
+
+    [JsonPropertyName("error_description")]
+    public string? ErrorDescription { get; init; }
+}
```
niemyjski: are these defined multiple times?? feels like I've seen these models twice. Also shouldn't they be in a models folder?

src/Exceptionless.Core/Services/OAuthClientMetadataService.cs
ejsmith: Yes, this path now caches fetch failures for 5 minutes per client metadata document, so repeated failures do not keep refetching. I think that is enough for this preview pass because successful documents are cached separately and DCR itself has an endpoint-specific throttle.

src/Exceptionless.Core/Services/OAuthClientMetadataService.cs
ejsmith: Done. This helper is now public static so the actual HTTP egress layer can reuse the URI validation / public-address checks.

src/Exceptionless.Core/Services/OAuthService.cs
ejsmith: Done. Refresh now takes a per-refresh-token lock before reading/disabling/minting, and the tests include concurrent refresh use where only one request succeeds.

src/Exceptionless.Web/Controllers/OAuthController.cs
ejsmith: Done. Added endpoint-specific DCR throttling and moved the limit to OAuthServer:DynamicClientRegistrationIpLimit so it can be tightened without a deploy. It still defaults to 20/hour/IP.

src/Exceptionless.Web/Mcp/ExceptionlessMcpTools.cs
ejsmith: Partially done in a low-risk way: MCP error contracts and response DTOs are now split into McpErrors.cs and McpModels.cs. I did not do the larger read/write/query-service split in this PR because it would be a much broader refactor of code that is now covered and passing; I would rather do that as a follow-up once this first hosted MCP server lands.

src/Exceptionless.Core/Services/OAuthService.cs
ejsmith: Done. The DCR per-IP limit is no longer a controller magic number; it is now OAuthServer:DynamicClientRegistrationIpLimit with the current default of 20.

src/Exceptionless.Core/Services/OAuthService.cs
ejsmith: I left this validation explicit for now. The service is validating protocol-level OAuth/DCR rules, not just persistence model shape, and keeping those checks near the OAuth flow makes the returned OAuth error codes/messages easier to control. We can revisit MiniValidator if this grows.

src/Exceptionless.Core/Services/OAuthService.cs
ejsmith: I left the explicit CreatedUtc/UpdatedUtc values here because the response includes client_id_issued_at and the tests assert the issued timestamp behavior. Using the same utcNow also keeps the persisted document and registration response deterministic.

src/Exceptionless.Core/Services/OAuthService.cs
ejsmith: I kept these as the OAuth system user id. DCR and client metadata discovery are machine/client registration flows, not user-owned admin edits, so recording them as system-created avoids attributing anonymous DCR writes to whichever user later authorizes the client.

src/Exceptionless.Core/Services/OAuthService.cs
ejsmith: I left this explicit for the same reason as the earlier validation comment: this is request/protocol validation that needs OAuth-specific error codes. Data annotations would still need mapping back into invalid_request, invalid_scope, invalid_redirect_uri, etc.

src/Exceptionless.Core/Services/OAuthService.cs
ejsmith: Handled by the refresh rotation path. Reuse now returns invalid_grant after the original refresh token has been disabled, and the concurrent-refresh test verifies only one caller can exchange the same refresh token.

src/Exceptionless.Web/ClientApp/src/lib/features/admin/api.svelte.ts
ejsmith: Agreed long-term, but I intentionally kept this initial management UI global-admin only. The hosted MCP path uses DCR/metadata so normal users do not need this page to connect Claude/Codex, and opening customer-managed OAuth apps needs product/permission design beyond this PR.

src/Exceptionless.Web/ClientApp/src/lib/features/admin/models.ts
ejsmith: Agreed in principle. I left the Svelte admin models hand-written for this PR to stay consistent with the current ClientApp pattern in this feature area. Switching this to generated OpenAPI models should be a separate frontend/API-client cleanup.

src/Exceptionless.Web/ClientApp/src/lib/features/admin/schemas.ts
ejsmith: Same as the model comment: likely worth doing once the ClientApp has a consistent generated-client story, but I did not want to introduce that tooling/change in this MCP PR.

src/Exceptionless.Web/Controllers/OAuthController.cs
ejsmith: I kept the explicit api/v2/oauth routes for now because these are public OAuth endpoints and changing controller-level routing would be easier to get subtly wrong. The current route shape is also what the metadata advertises and what Claude/Codex have been tested against.

src/Exceptionless.Core/Bootstrapper.cs
```diff
@@ -182,13 +185,45 @@ public static void RegisterServices(IServiceCollection services, AppOptions appO
...
+            {
+                await socket.ConnectAsync(new IPEndPoint(address, context.DnsEndPoint.Port), cancellationToken);
+                return new NetworkStream(socket, ownsSocket: true);
+            }
+            catch (Exception ex)
+            {
+                lastException = ex;
+                socket.Dispose();
+            }
+        }
```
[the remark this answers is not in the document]
ejsmith: Leaving this as-is. We intentionally did not take the LINQ rewrite; the current loop keeps the socket lifecycle and skip/connect behavior straightforward without adding a LINQ dependency for a security-sensitive egress path.

PR #2321, opened 2026-06-22
Add event version column

### LINKED ISSUE #1299: UI: Allow customizing event columns
It has been requested to be able to show specific columns like version in events grid.

ejsmith:
## Summary
- Adds `version` to event summary responses from the existing event `@version` metadata.
- Adds a hidden-by-default `Version` column to the Svelte event table so it can be selected and saved in saved views.

Closes #1299.

## Verification
- `npm run check` in `src/Exceptionless.Web/ClientApp`
- `dotnet test tests/Exceptionless.Tests/Exceptionless.Tests.csproj -- --filter-class Exceptionless.Tests.Plugins.SummaryDataTests`


src/Exceptionless.Core/Plugins/Formatting/FormattingPluginManager.cs
```diff
@@ -55,6 +59,15 @@ public SummaryData GetEventSummaryData(PersistentEvent ev)
         throw new InvalidOperationException();
     }
 
+    private static void AddVersionSummaryData(SummaryData summary, PersistentEvent ev)
+    {
+        string? version = ev.GetVersion();
+        if (String.IsNullOrWhiteSpace(version) || summary.Data is not IDictionary<string, object?> data)
+            return;
+
+        data.TryAdd("Version", version);
```
ejsmith: Why do we need to add the version to the data instead of just accessing it where it is?

src/Exceptionless.Web/ClientApp/src/lib/features/events/components/table/options.svelte.ts
```diff
@@ -102,6 +103,16 @@ export function getColumns<TSummaryModel extends SummaryModel<SummaryTemplateKey
                     class: 'w-28'
                 }
             },
+            {
+                accessorKey: nameof<EventSummaryModel<SummaryTemplateKeys>>('version'),
+                cell: (prop) => formatTextColumn(prop.getValue()),
+                enableSorting: false,
+                header: 'Version',
+                id: 'version',
```
ejsmith, line 111: This new column id also needs to be added to the saved-view validation allow-list. The saved-view picker sends `columns` and `column_order` when saving/updating a view, but `NewSavedView.ValidColumnIds` currently allows events/stream columns only through `summary`, `user`, `date`, `message`, `type`, `exception_type`, `source`, `name`, and `level`. If a user enables or reorders `version`, the API will reject the saved view with 422. Please add `version` to the `events` and `stream` allow-lists and cover that in the saved-view column validation tests.

ejsmith (commented):
I found one blocker.

- The new `version` table column can be enabled in the UI, but saved-view validation still rejects it. Save/update sends the current `columns` and `column_order` to the saved-view endpoints, and `NewSavedView.ValidColumnIds` does not include `version` for `events` or `stream`, so a user who enables or orders this column will get a 422 when saving/updating the view. Add `version` to the event and stream allow-lists and cover it in the saved-view validation tests.
- CI note: the current `test-api` failure is from the branch being behind `main`; the failing paging tests are covered by `Fix event cursor pagination links (#2322)`. Rebase/merge `main` and rerun checks after the allow-list fix.

src/Exceptionless.Web/ClientApp/src/lib/features/events/components/table/options.svelte.ts
ejsmith: Fixed in c0986269a by adding `version` to the `events` and `stream` saved-view validation allow-lists and adding controller coverage for saving it through both `columns` and `column_order`.

src/Exceptionless.Core/Plugins/Formatting/FormattingPluginManager.cs
ejsmith: Resolved by the current shape: the column reads from top-level `EventSummaryModel.Version` / `row.version`, so we are no longer duplicating the version into summary data.

ejsmith:
Addressed the remaining review feedback in c0986269a:

- Added `version` to the `events` and `stream` saved-view validation allow-lists.
- Added `SavedViewControllerTests.PostAsync_VersionColumnForEventViews_Succeeds` coverage for saving `version` in both `columns` and `column_order`.
- Replied on the older data-shape thread; the column now uses top-level `EventSummaryModel.Version` / `row.version` instead of duplicating into summary data.

Local verification: `dotnet test tests\Exceptionless.Tests\Exceptionless.Tests.csproj -- --filter-class Exceptionless.Tests.Controllers.SavedViewControllerTests` passed, 92 tests.

Current GitHub checks for c0986269a are still running: `version`, `test-api`, and `test-client`.

PR #2410, opened 2026-07-28
Repair invalid verified user email state

niemyjski:
## Summary

- add a versioned, resumable migration that repairs verified users retaining stale email-verification tokens or expirations
- preserve verified status and unrelated user fields while invalidating only user and email cache entries
- add detailed migration logging, failure guards, post-migration validation, and idempotent behavior
- cover the historical corrupted record shape, partial corruption, cached reads, reruns, and unaffected valid users

## Why

Older authentication flows could persist users as verified while leaving email-verification credentials populated. Current validation correctly rejects that contradictory state, which causes unrelated full-user saves such as forgot-password to return 422. The producer has already been corrected; this migration repairs existing records without weakening the invariant or forcing users through verification again.

## Verification

- dotnet build tests/Exceptionless.Tests/Exceptionless.Tests.csproj --no-restore
- dotnet test tests/Exceptionless.Tests/Exceptionless.Tests.csproj --no-build -- --filter-namespace Exceptionless.Tests.Migrations (8 passed)
- scoped dotnet format --verify-no-changes for all touched files
- git diff --cached --check

## Breaking changes

None.


src/Exceptionless.Core/Migrations/003_RepairVerifiedUserEmailVerification.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        _logger.LogInformation(
+            "Starting verified user email verification repair migration. Version={MigrationVersion} Index={Index} Query={Query}",
+            Version,
+            index,
+            INVALID_VERIFIED_USER_DESCRIPTION);
+
+        _logger.LogInformation("Refreshing user index {Index} before counting invalid verified user records", index);
+        var refreshResponse = await _client.Indices.RefreshAsync(index, context.CancellationToken);
+        _logger.LogRequest(refreshResponse, LogLevel.Information);
```
niemyjski:
```suggestion
        _logger.LogRequest(refreshResponse);
```

Default consistency is immediate so we don't need to do anything here.

src/Exceptionless.Core/Migrations/003_RepairVerifiedUserEmailVerification.cs
```diff
@@ -0,0 +1,167 @@
...
+                index,
+                stopwatch.Elapsed);
+            return;
+        }
+
+        _logger.LogWarning(
+            "Repairing {RecordsToRepair:N0} verified user record(s) by removing verify_email_address_token and verify_email_address_token_expiration. Verified status and all unrelated fields will be preserved",
+            recordsToRepair);
+
+        var updateResponse = await _client.UpdateByQueryAsync<User>(
```
niemyjski: I'd rather use the repositories call to query and find users and not use raw elastic client.

src/Exceptionless.Core/Migrations/003_RepairVerifiedUserEmailVerification.cs
```diff
@@ -0,0 +1,167 @@
...
+public sealed class RepairVerifiedUserEmailVerification : MigrationBase
+{
+    private const string INVALID_VERIFIED_USER_DESCRIPTION = "is_email_address_verified=true and either a verify email address token or a non-default token expiration is present";
+    private const string REMOVE_VERIFICATION_FIELDS_SCRIPT = "ctx._source.remove('verify_email_address_token'); ctx._source.remove('verify_email_address_token_expiration');";
+
+    private readonly ICacheClient _cache;
+    private readonly ElasticsearchClient _client;
+    private readonly ExceptionlessElasticConfiguration _configuration;
+
+    public RepairVerifiedUserEmailVerification(ExceptionlessElasticConfiguration configuration, ILoggerFactory loggerFactory) : base(loggerFactory)
```
niemyjski: This should be suffixed with Migration..

tests/Exceptionless.Tests/Migrations/RepairVerifiedUserEmailVerificationMigrationTests.cs
```diff
@@ -0,0 +1,216 @@
...
+    protected override void RegisterServices(IServiceCollection services)
+    {
+        services.AddTransient<RepairVerifiedUserEmailVerification>();
+        services.AddSingleton<ILock>(EmptyLock.Empty);
+        base.RegisterServices(services);
+    }
+
+    [Fact]
+    public async Task RunAsync_InvalidVerifiedUser_RepairsVerificationStateAndInvalidatesCache()
+    {
```
niemyjski, line 32: missing // Arrange, Act, Assert, check entire pr.

src/Exceptionless.Core/Migrations/003_RepairVerifiedUserEmailVerification.cs
```diff
@@ -0,0 +1,167 @@
...
+            index);
+
+        if (recordsToRepair is 0)
+        {
+            _logger.LogInformation(
+                "Verified user email verification repair migration completed without changes. Version={MigrationVersion} Index={Index} Duration={Duration}",
+                Version,
+                index,
+                stopwatch.Elapsed);
+            return;
```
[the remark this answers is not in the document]
niemyjski: Fixed. Cache-prefix invalidation now runs unconditionally before post-migration verification, including when a resumable retry finds zero invalid documents. Added RunAsync_NoInvalidUsers_InvalidatesStaleUserCache to reproduce and protect this path.

src/Exceptionless.Core/Migrations/003_RepairVerifiedUserEmailVerification.cs
niemyjski: Fixed. The raw Elasticsearch refresh and request logging were removed. The migration now uses IUserRepository, whose default consistency is already immediate.

src/Exceptionless.Core/Migrations/003_RepairVerifiedUserEmailVerification.cs
niemyjski: Fixed. Count, patch, and post-migration verification now use IUserRepository with one shared typed predicate. The patch uses the canonical MarkEmailAddressVerified helper through ActionPatch.

src/Exceptionless.Core/Migrations/003_RepairVerifiedUserEmailVerification.cs
niemyjski: Fixed. Renamed the production type and file to RepairVerifiedUserEmailVerificationMigration.

tests/Exceptionless.Tests/Migrations/RepairVerifiedUserEmailVerificationMigrationTests.cs
niemyjski: Fixed. Added explicit Arrange, Act, and Assert sections throughout the migration integration tests and checked every test in the PR.

src/Exceptionless.Core/Repositories/Interfaces/IUserRepository.cs
```diff
@@ -11,4 +11,5 @@ public interface IUserRepository : ISearchableRepository<User>
     Task<User?> GetUserByOAuthProviderAsync(string provider, string providerUserId);
     Task<User?> GetByVerifyEmailAddressTokenAsync(string token);
     Task<FindResults<User>> GetByOrganizationIdAsync(string organizationId, CommandOptionsDescriptor<User>? options = null);
+    Task<FindResults<User>> GetVerifiedUsersWithStaleVerificationDataAsync(CommandOptionsDescriptor<User>? options = null);
```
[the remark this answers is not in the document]
niemyjski: Fixed in 784a5b28f. Removed the migration-only method from IUserRepository and UserRepository, restoring the shared repository API exactly. The one-off typed query now lives in a private FindUsersToRepairAsync helper inside the migration and uses the existing inherited FindAsync API.

src/Exceptionless.Core/Migrations/003_RepairVerifiedUserEmailVerificationMigration.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        do
+        {
+            context.CancellationToken.ThrowIfCancellationRequested();
+            batch++;
+
+            foreach (var user in users.Documents)
+                user.MarkEmailAddressVerified();
+
+            await _userRepository.SaveAsync(users.Documents);
```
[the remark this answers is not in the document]
niemyjski: Fixed in a45da7247. The migration now applies one strongly typed bulk ActionPatch<User> to the IDs from each search page instead of saving the stale search snapshots. The patch only invokes MarkEmailAddressVerified, preserves unrelated concurrent fields, retains normal notifications, and uses document-based cache invalidation.

src/Exceptionless.Core/Migrations/003_RepairVerifiedUserEmailVerificationMigration.cs
```diff
@@ -0,0 +1,88 @@
...
+        int batch = 0;
+        var users = await FindUsersToRepairAsync();
+
+        if (users.Documents.Count == 0)
+        {
+            _logger.LogInformation(
+                "Verified user email verification repair migration completed without changes. Version={MigrationVersion} Duration={Duration}",
+                Version,
+                stopwatch.Elapsed);
+            return;
```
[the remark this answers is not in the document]
niemyjski: No broad prefix eviction added. The migration now uses ActionPatch<User>, whose repository contract performs document-based invalidation and passes each modified User through UserRepository.InvalidateCacheAsync, clearing both ID and custom email cache entries. The integration test primes both keys and verifies the migration invalidates them. A process terminating between datastore commit and repository invalidation is a system-wide cache atomicity concern; this one-off migration should not clear every user/email cache on a zero-result run to compensate for that hypothetical window.

src/Exceptionless.Core/Migrations/003_RepairVerifiedUserEmailVerificationMigration.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        do
+        {
+            context.CancellationToken.ThrowIfCancellationRequested();
+            batch++;
+
+            await _userRepository.PatchAsync(
+                users.Documents.Select(user => user.Id).ToArray(),
+                new ActionPatch<User>(user => user.MarkEmailAddressVerified()));
```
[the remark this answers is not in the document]
niemyjski: Fixed in 781dbbfe5. The inline ActionPatch<User> now rechecks the current document and no-ops unless it is still verified and still contains stale verification data. If an email update made the account unverified after the search, the patch returns false and preserves the newly issued token and expiration.

PR #2436, opened 2026-08-04
Fix long tag overflow popover interactions

niemyjski:
## Summary

- Keep complete long-tag values available from visible tags and `+N` overflow summaries.
- Preserve filter and Alt/Option-copy actions for clickable hidden tags.
- Bound and wrap popovers on narrow viewports without changing desktop rendering.
- Add realistic synthetic Storybook fixtures and focused regression coverage.
- Make Aspire, Storybook, and Lume docs Run actions invoke tracked executable scripts atomically.

## Root cause

Hidden tags were rendered through the normal tooltip path inside the `+N` tooltip. Moving from the overflow trigger to a hidden tag opened or closed nested tooltip state, so the full hidden value was not reliably viewable or actionable. The overflow path now renders bounded, wrapping values directly and reuses the existing click handler.

The visible badge is an `inline-flex` container. Truncation is applied and measured on its inner text span so padding and flex layout do not distort the overflow check.

The first Aspire wrapper was not an Aspire crash: its historical log reached a healthy dashboard and running Api/App/Elasticsearch/Jobs before receiving a normal termination signal. The current desktop failure was separate: multiline TOML commands were pasted into an interactive zsh terminal, leaving the shell at `then>` before Aspire could run. The Run actions now point to executable project-local scripts. The Aspire script resolves the worktree without requiring `CODEX_WORKTREE_PATH`, builds the AppHost from the repository root, starts Aspire with `--no-build`, and uses `exec` so logs and Ctrl-C remain attached.

## Verification

- Focused TagList tests — 7 passed, including hidden filtering and Alt/Option-copy behavior
- `npm run validate` — formatting, Svelte check, Prettier, and ESLint passed; 0 Svelte errors or warnings
- `npm run build-storybook` — completed successfully
- Storybook browser matrix — read-only, clickable, overflow, keyboard focus, extreme values, and 320px viewport passed with no browser errors
- Exact executable Storybook action with `CODEX_WORKTREE_PATH` absent — stayed attached; Storybook and the long-tag story URL both returned HTTP 200
- Exact executable Aspire action with `CODEX_WORKTREE_PATH` absent — AppHost build succeeded, dashboard stayed attached, Api/App/Jobs and infrastructure reported healthy, API and Svelte app returned HTTP 200, and Ctrl-C stopped cleanly
- Environment TOML parsed; the three executable run scripts passed `bash -n`; `git diff --check` passed

## Scope

This pull request contains the TagList component, stories, focused tests, and the narrowly scoped environment action repair. It does not include the earlier custom background process manager or a redundant `Run Site` action; `Run Aspire` starts the full local site. Local screenshots and dogfood reports remain ignored and are not part of the pull request.

## Before / after

The before/after screenshots are embedded here as GitHub user attachments; they are not repository files.


niemyjski:
@codex review

ejsmith (approved):

PR #2460, opened 2026-08-12
Fix Intercom connectivity and update noise

ejsmith:
## Summary

- allow Intercom's current US API, upload, and realtime endpoints in the web CSP
- remove the periodic Intercom update loop and avoid duplicate updates during initial boot
- centralize Intercom teardown for manual logout, automatic `401` logout, and OAuth reauthentication
- clear residual `intercom-*` cookies even when tracking prevention blocks the SDK
- update Intercom only for normalized SvelteKit route changes or changed identity/company data
- send identity plus `last_request_at` for route-only updates instead of the full boot payload
- cover boot, navigation, identity refresh, teardown, cookie cleanup, throttling safeguards, and delayed SDK initialization

## Why

Production browsers blocked dynamically selected Intercom realtime WebSocket hosts because the CSP allowed only one legacy endpoint. The Svelte client also called Intercom update every 90 seconds, consuming the documented 20-calls-per-30-minute quota and repeatedly generating third-party traffic and console noise.

Intercom session cleanup was only explicit during manual logout. Automatic authentication loss relied on a component effect that could be unmounted before running, and `shutdown` alone can leave Intercom's anonymous identifier cookie behind. Session teardown now runs from every Svelte authentication-ending path and clears all visible Intercom cookies.

Normalized route IDs prevent navigation between individual events or stacks from consuming additional update calls. Meaningful route changes send only identity and the current request timestamp, while user or company changes send only the fields that changed.

The legacy Angular application remains isolated from the Svelte lifecycle changes. The shared CSP change preserves every previous source exactly and only adds Intercom connection destinations.

## Validation

- `npm run validate` in `ClientApp`
- `npm run test:unit` in `ClientApp` — 46 files, 464 tests passed
- `npm run build` in `ClientApp`
- `npm run lint` in `ClientApp.angular`
- `npm run build` in `ClientApp.angular`
- `dotnet msbuild src/Exceptionless.Web/Exceptionless.Web.csproj /t:Build /p:Restore=false /m:1 /v:minimal`

## Breaking changes

None.

### DIFF
--- a/src/Exceptionless.Web/ClientApp/src/lib/features/auth/api.svelte.ts
+++ b/src/Exceptionless.Web/ClientApp/src/lib/features/auth/api.svelte.ts
@@ -2,13 +2,13 @@ import { env } from '$env/dynamic/public';
 import { getIntercomTokenSessionKey, intercomTokenRefreshIntervalMs } from '$features/intercom/config';
 import { organization } from '$features/organizations/context.svelte';
 import { ProblemDetails, useFetchClient } from '@foundatiofx/fetchclient';
-import { hide as hideIntercom, shutdown as shutdownIntercom } from '@intercom/messenger-js-sdk';
 import { createQuery, type QueryClient } from '@tanstack/svelte-query';
 
 import type { Login, TokenResult } from './models';
 
 import { endSession } from './exceptionless-session';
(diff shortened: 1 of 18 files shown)


ejsmith:
@codex review

PR #2466, opened 2026-08-13
Prevent transient Service Status flicker

ejsmith:
## Summary

- confirm transient API failures against the same-origin health endpoint before leaving the current page
- coalesce concurrent health checks and Service Status navigation, while preserving the complete return URL
- make WebSocket callbacks socket-specific and suppress errors from intentional shutdowns
- add unit and Playwright regressions for healthy and unavailable service paths

## Root cause

Every API response with status 0, 408, or 503 immediately navigated to Service Status. Several responses during rapid navigation could therefore repeatedly mount and unmount the authenticated layout. Each teardown closed the push socket, and a socket still connecting reported that intentional interruption through the normal error path.

## Behavior and compatibility

An isolated transient API response no longer interrupts navigation when `/health` remains healthy. If the health probe also fails, the app performs one coalesced Service Status redirect and retains the current path, query string, and hash for recovery.

Existing API contracts and push message formats are unchanged. Breaking changes: none.

## Verification

- `npm run validate` — 0 Svelte errors and 0 warnings
- `npm run test:unit` — 48 files, 486 tests passed
- `npm run build` — passed
- focused localhost Playwright — new Service Status regression plus event and stack effect-chaos coverage, 3 tests passed
- controlled localhost reproduction — healthy service: 1 health probe, 0 status redirects, 0 WebSocket errors; unavailable service: 1 status redirect with the exact return URL
- no `effect_update_depth_exceeded`, maximum-update-depth, console, or page errors in the focused browser suite

### DIFF
--- a/src/Exceptionless.Web/ClientApp/e2e/tests/service-status-navigation.e2e.ts
+++ b/src/Exceptionless.Web/ClientApp/e2e/tests/service-status-navigation.e2e.ts
@@ -1,0 +1,71 @@
+import { expect, test } from '../fixtures/e2e-test';
+
+test('transient API failures only open Service Status when the health probe fails', async ({ e2eScenario, page }) => {
+    const webSocketErrors: string[] = [];
+    const serviceStatusNavigations: string[] = [];
+    let healthRequests = 0;
+
+    page.on('console', (message) => {
+        if (message.type() === 'error' && message.text().includes('[WebSocketClient]')) {
+            webSocketErrors.push(message.text());
+        }
+    });
+    page.on('framenavigated', (frame) => {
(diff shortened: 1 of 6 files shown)


niemyjski:
Thermo-nuclear code-quality pass found one straightforward structural simplification, fixed in 5cf3b0e08: the redirector now coalesces the entire health-check/redirect operation behind one promise instead of coordinating separate health and navigation promises. This preserves the cache and concurrency behavior while removing state and the interleaving between phases.\n\nFocused verification on the exact PR head plus the fix:\n- `service-status-redirect.test.ts`: 5 passed\n- Prettier check: passed\n- `git diff --check`: passed\n\nThe hosted replacement CI run is the authoritative full-suite gate.

PR #2475, opened 2026-08-14
Allow deletion of legacy suspended organizations

niemyjski:
## What changed

Migrate the exact legacy Stripe suspension marker before normal organization writes validate it.

## Evidence

- Git history shows the old Stripe event handler wrote the literal `"Stripe"` into `SuspendedByUserId` in both suspension paths.
- A later Stripe billing change introduced the all-zero 24-character system user ID used today.
- Existing `SystemUserId` constants for OAuth and predefined saved views are different IDs, so they are not interchangeable.

## Fix

- Centralize the Stripe-specific legacy and current IDs.
- Add version 7 as a versioned, resumable migration.
- Scan suspended organizations, including soft-deleted records, through the existing typed organization repository.
- Filter the unmapped legacy field from loaded source and replace only the exact legacy marker.
- Refresh the current organization index after writes so reruns and subsequent API requests observe the migration.
- Preserve suspension metadata so deleting and restoring an organization does not silently clear its suspension state.

## Verification

- Five integration tests cover historical conversion, non-target records, soft-deleted records, idempotent reruns, and more than one migration page.
- Organization endpoint regression coverage verifies suspension state survives delete and restore.
- Migration registration test verifies version 7 and VersionedAndResumable registration.
- dotnet build Exceptionless.slnx --no-restore -m:1
- Hosted Build run 31840728876 passed all required checks.

No production identifiers or customer data are included.

### DIFF
--- a/src/Exceptionless.Core/Billing/StripeConstants.cs
+++ b/src/Exceptionless.Core/Billing/StripeConstants.cs
@@ -2,6 +2,5 @@ namespace Exceptionless.Core.Billing;
 
 internal static class StripeConstants
 {
-    public const string LegacySystemUserId = "Stripe";
     public const string SystemUserId = "000000000000000000000000";
 }
--- a/src/Exceptionless.Core/Billing/StripeConstants.cs
+++ b/src/Exceptionless.Core/Billing/StripeConstants.cs
@@ -1,0 +1,7 @@
+namespace Exceptionless.Core.Billing;
+
+internal static class StripeConstants
+{
+    public const string LegacySystemUserId = "Stripe";
+    public const string SystemUserId = "000000000000000000000000";
+}
--- a/tests/Exceptionless.Tests/Api/Endpoints/OrganizationEndpointTests.cs
+++ b/tests/Exceptionless.Tests/Api/Endpoints/OrganizationEndpointTests.cs
@@ -1,12 +1,10 @@
 using System.Net;
-using Elastic.Clients.Elasticsearch;
 using Exceptionless.Core;
 using Exceptionless.Core.Billing;
 using Exceptionless.Core.Extensions;
 using Exceptionless.Core.Models;
 using Exceptionless.Core.Models.Billing;
 using Exceptionless.Core.Repositories;
-using Exceptionless.Core.Repositories.Configuration;
 using Exceptionless.Core.Services;
 using Exceptionless.Core.Utility;
 using Exceptionless.Tests.Extensions;
@@ -24,8 +22,8 @@ namespace Exceptionless.Tests.Api.Endpoints;
 
 public sealed class OrganizationEndpointTests : IntegrationTestsBase
 {
+    private const string ValidSuspendedByUserId = "660000000000000000000001";
     private readonly IOrganizationRepository _organizationRepository;
-    private readonly ExceptionlessElasticConfiguration _configuration;
     private readonly IProjectRepository _projectRepository;
     private readonly IUserRepository _userRepository;
     private readonly BillingManager _billingManager;
@@ -35,7 +33,6 @@ public sealed class OrganizationEndpointTests : IntegrationTestsBase
     public OrganizationEndpointTests(ITestOutputHelper output, AppWebHostFactory factory) : base(output, factory)
     {
         _organizationRepository = GetService<IOrganizationRepository>();
-        _configuration = GetService<ExceptionlessElasticConfiguration>();
         _projectRepository = GetService<IProjectRepository>();
         _userRepository = GetService<IUserRepository>();
         _billingManager = GetService<BillingManager>();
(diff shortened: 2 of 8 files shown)


tests/Exceptionless.Tests/Api/Endpoints/OrganizationEndpointTests.cs
```diff
@@ -22,6 +22,7 @@ namespace Exceptionless.Tests.Api.Endpoints;
 
 public sealed class OrganizationEndpointTests : IntegrationTestsBase
 {
+    private const string ValidSuspendedByUserId = "660000000000000000000001";
```
niemyjski: This should reuse the stripe constant.

src/Exceptionless.Core/Billing/StripeConstants.cs
```diff
@@ -0,0 +1,7 @@
+namespace Exceptionless.Core.Billing;
+
+internal static class StripeConstants
+{
+    public const string LegacySystemUserId = "Stripe";
```
niemyjski: this belongs in the migration.

src/Exceptionless.Core/Billing/StripeConstants.cs
```diff
@@ -0,0 +1,7 @@
...
+    public const string SystemUserId = "000000000000000000000000";
```
niemyjski, line 5: is this a valid object id?

tests/Exceptionless.Tests/Migrations/MigrateLegacyStripeSuspensionUserIdMigrationTests.cs
```diff
@@ -0,0 +1,233 @@
...
+using Foundatio.Utility;
+using Xunit;
+
+namespace Exceptionless.Tests.Migrations;
+
+public sealed class MigrateLegacyStripeSuspensionUserIdMigrationTests : IntegrationTestsBase
+{
+    // This is the exact value written by StripeEventHandler before the system user id was introduced.
+    private const string HistoricalStripeSuspensionUserId = "Stripe";
+    private const string ValidSuspendedByUserId = "660000000000000000000001";
```
niemyjski: reuse known const

tests/Exceptionless.Tests/Migrations/MigrateSavedViewColumnsIntegrationTests.cs
```diff
@@ -95,8 +95,8 @@ public async Task DataSeedStartupAction_OnlyRepeatableMigrationsPending_SeedsDat
         var migrationStateRepository = GetService<IMigrationStateRepository>();
         await migrationStateRepository.AddAsync(new MigrationState
         {
-            Id = "5",
-            Version = 5,
+            Id = "7",
+            Version = 7,
```
niemyjski, line 100: Why did this change?

tests/Exceptionless.Tests/Api/Endpoints/OrganizationEndpointTests.cs
niemyjski: Done in 18bf6d23d: the organization delete/restore regression now reuses StripeConstants.SystemUserId instead of introducing another synthetic ID.

src/Exceptionless.Core/Billing/StripeConstants.cs
niemyjski: Yes. ObjectIdAttribute defines a 24-character hexadecimal object ID, and 000000000000000000000000 is exactly 24 hexadecimal characters. This is also the existing Stripe system identity used by the pre-change StripeEventHandler.

src/Exceptionless.Core/Billing/StripeConstants.cs
niemyjski: Done in 18bf6d23d: the historical marker is no longer part of StripeConstants. It now belongs to the migration as MigrateLegacyStripeSuspensionUserId.LegacySystemUserId, and the migration tests reuse it.

tests/Exceptionless.Tests/Migrations/MigrateSavedViewColumnsIntegrationTests.cs
niemyjski: Version 7 is required here because the new migration is versioned. DataSeedService will not seed repeatable data while a versioned migration is pending; the fixture must mark the current version, 7, complete to isolate the repeatable-seed behavior. I added a comment documenting that contract.

tests/Exceptionless.Tests/Migrations/MigrateLegacyStripeSuspensionUserIdMigrationTests.cs
niemyjski: Done in 18bf6d23d: migration tests now reuse MigrateLegacyStripeSuspensionUserId.LegacySystemUserId for the historical marker and TestConstants.UserId for the known valid non-Stripe user ID.

PR #2491, opened 2026-08-18
Remove duplicate pipeline performance test

ejsmith:
## What changed

- remove `PipelinePerformanceAsync` from the required backend suite
- remove its now-unused `System.Diagnostics` import

## Why

`PipelinePerformanceAsync` iterated every file in `Pipeline/Data`, ran each batch through `EventPipeline`, and repeated the same processed/cancelled/error assertions already made by the adjacent `ProcessEventsAsync` theory for every one of those files. Its only additional behavior was logging elapsed time; it had no performance threshold and could not detect a performance regression.

This removes duplicate execution without removing fixture or behavioral coverage. Actual performance tracking should live in an explicit benchmark or scheduled performance job with a threshold and retained history.

## Measured effect

- the duplicate test took about 11.1 seconds in the local full-suite TRX
- the focused class now passes its remaining 56 cases in 12.9 seconds of test execution

## Verification

- `dotnet test --project tests/Exceptionless.Tests/Exceptionless.Tests.csproj --configuration Release -- --filter-class Exceptionless.Tests.Pipeline.EventPipelineTests`
- 55 passed, 1 skipped, 0 failed

## Breaking changes

None.

### DIFF
--- a/tests/Exceptionless.Tests/Pipeline/EventPipelineTests.cs
+++ b/tests/Exceptionless.Tests/Pipeline/EventPipelineTests.cs
@@ -1,4 +1,3 @@
-using System.Diagnostics;
 using System.Globalization;
 using System.Text;
 using Exceptionless.Core.Billing;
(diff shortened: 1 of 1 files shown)


ejsmith:
Closing this isolated experiment because it did not produce a material, repeatable build-time improvement.

The removed test took 11.1 seconds locally and duplicated the fixture coverage and processed/cancelled/error assertions in `ProcessEventsAsync`; its only unique output was an elapsed-time log with no threshold.

Hosted `.NET Tests with Coverage` samples:

| Attempt | Test step |
| --- | ---: |
| 1 | 7m47s |
| 2 | 9m19s |
| 3 | 9m23s |

The 9m19s median is only two seconds below the recent 9m21s baseline. The first sample was runner variance, not a credible consequence of deleting an 11-second local test. This is below the 30-second/5% retention gate, so the PR is intentionally closed.

PR #2593, opened 2026-09-21
Fix cross-project bot throttling and event cleanup

landong-coder:
Bot throttling currently shares an IP/time-window counter across projects and organizations. Its cleanup job also deletes matching events across the entire organization, so one project's traffic can throttle another project or remove events from a project with bot filtering disabled.

Scope counters and batch grouping by organization, project, and client IP. Cache keys use `Organization:{organizationId}:Project:{projectId}:bot:{period}:{ip}` through a dedicated helper. Cleanup calls a new repository overload with project ID as the second parameter; the existing organization-wide overload remains available. Reject cleanup tasks with missing scope or IP information before deletion.

### Tests and validation
- Rebased onto upstream `main` at `deffcf185`.
- The plugin tests now inherit the repository's existing `TestWithLoggingBase`, use its logger factory, and override asynchronous disposal.
- Eight focused unit tests passed, covering cross-scope counters, mixed batches, window rollover, cleanup argument forwarding, and missing project IDs.
- Added four repository integration cases verifying actual deletion preserves other projects, organizations, IPs, and records outside the date range, including optional date bounds.
- The integration cases compile, but local execution is blocked during fixture initialization: `Aspire.Hosting.DistributedApplicationException: Container runtime 'docker' could not be found.` No deletion assertions ran locally. Hosted CI is pending.
- `git diff --check` passed.

### Compatibility and deployment
No HTTP API, configuration-key, or work-item payload changes. The existing repository overload is preserved, with an additional project-scoped overload. Existing queued cleanup tasks with valid project IDs remain supported; invalid tasks fail before deletion.

The new cache keys start fresh per-project counters; existing shared keys expire naturally. Update all event-processing and cleanup worker instances together, since old workers retain the previous behavior.


src/Exceptionless.Core/Jobs/WorkItemHandlers/RemoveBotEventsWorkItemHandler.cs
```diff
@@ -27,12 +30,21 @@ public RemoveBotEventsWorkItemHandler(IEventRepository eventRepository, ILockPro
...
+        ArgumentException.ThrowIfNullOrWhiteSpace(wi.OrganizationId);
+        ArgumentException.ThrowIfNullOrWhiteSpace(wi.ProjectId);
+        ArgumentException.ThrowIfNullOrWhiteSpace(wi.ClientIpAddress);
 
         using var _ = Log.BeginScope(new ExceptionlessState().Organization(wi.OrganizationId).Project(wi.ProjectId).Tag("Delete").Tag("Bot"));
         Log.LogInformation("Received remove bot events work item OrganizationId={OrganizationId} ProjectId={ProjectId}, ClientIpAddress={ClientIpAddress}, UtcStartDate={UtcStartDate}, UtcEndDate={UtcEndDate}", wi.OrganizationId, wi.ProjectId, wi.ClientIpAddress, wi.UtcStartDate, wi.UtcEndDate);
 
         await context.ReportProgressAsync(0, $"Starting deleting of bot events... OrganizationId={wi.OrganizationId}");
-        long deleted = await _eventRepository.RemoveAllAsync(wi.OrganizationId, wi.ClientIpAddress, wi.UtcStartDate, wi.UtcEndDate);
+        var query = new RepositoryQuery<PersistentEvent>()
```
niemyjski: I think this would be better to update this RemoveAllAsync overload to take a project id as a second parameter because then we can test just this in integration tests to verify the parameter / mock it out. This would also save changes here, the argument changes are great!

src/Exceptionless.Core/Plugins/EventProcessor/Default/0_ThrottleBotsPlugin.cs
```diff
@@ -33,19 +33,25 @@ public override async Task EventBatchProcessingAsync(ICollection<EventContext> c
...
-            if (String.IsNullOrEmpty(clientIpAddressGroup.Key) || clientIpAddressGroup.Key.IsPrivateNetwork())
+            var scope = clientIpAddressGroup.Key;
+            if (String.IsNullOrEmpty(scope.ClientIpAddress) || scope.ClientIpAddress.IsPrivateNetwork())
+            {
                 continue;
+            }
 
             var clientIpContexts = clientIpAddressGroup.ToList();
-            string throttleCacheKey = String.Concat("bot:", clientIpAddressGroup.Key, ":", _timeProvider.GetUtcNow().UtcDateTime.Floor(_throttlingPeriod).Ticks);
+            string throttleCacheKey = $"bot:{scope.OrganizationId}:{scope.ProjectId}:{scope.ClientIpAddress}:{_timeProvider.GetUtcNow().UtcDateTime.Floor(_throttlingPeriod).Ticks}";
```
niemyjski: I think we should move this to a helper and scope the cache key following this pattern:

This can be a method or the following (we've added AgumentException.ThrowIf... in some cases). The reason behind this is it adds logical namespaces that you can view as a tree view in most clients and allows us to do wild card cache cleanup for a given org / project.

private static string CacheKey(string organizationId, string projectId, string clientIpAddress, period) => String.Concat("Organization:", organizationId, ":Project:", projectId, ":bot:", period, ":", clientIpAddress);

tests/Exceptionless.Tests/Plugins/ThrottleBotsPluginTests.cs
```diff
@@ -0,0 +1,95 @@
...
+namespace Exceptionless.Tests.Plugins;
+
+public sealed class ThrottleBotsPluginTests : IDisposable
+{
+    private readonly FakeTimeProvider _time = new(new DateTimeOffset(2026, 9, 21, 2, 25, 0, TimeSpan.Zero));
+    private readonly InMemoryCacheClient _cache;
+    private readonly InMemoryQueue<WorkItemData> _queue = new();
+    private readonly ThrottleBotsPlugin _plugin;
+
+    public ThrottleBotsPluginTests()
```
niemyjski: If we derive from UnitTestBase class we get logging and much more for free and then we can also override for async disposable.

niemyjski (commented):
Thanks a ton for your pr, these changes would really help a lot and we really appreciate it. I left a few comments which I feel would really help testing and debugging this in production scenarios.

src/Exceptionless.Core/Repositories/Interfaces/IEventRepository.cs
```diff
@@ -12,6 +12,7 @@ public interface IEventRepository : IRepositoryOwnedByOrganizationAndProject<Per
     Task<FindResults<PersistentEvent>> GetOpenSessionsAsync(DateTime createdBeforeUtc, CommandOptionsDescriptor<PersistentEvent>? options = null);
     Task<bool> UpdateSessionStartLastActivityAsync(string id, DateTime lastActivityUtc, bool isSessionEnd = false, bool hasError = false, bool sendNotifications = true);
     Task<long> RemoveAllAsync(string organizationId, string? clientIpAddress, DateTime? utcStart, DateTime? utcEnd, CommandOptionsDescriptor<PersistentEvent>? options = null);
+    Task<long> RemoveAllAsync(string organizationId, string projectId, string clientIpAddress, DateTime? utcStart, DateTime? utcEnd, CommandOptionsDescriptor<PersistentEvent>? options = null);
```
ejsmith: Please rename both the existing organization-scoped method and the new project-scoped method so their deletion scope is explicit and the naming is consistent. For example:

- Existing: `RemoveAllByOrganizationAndClientIpAsync`
- New: `RemoveAllByProjectAndClientIpAsync`

The current overloads resolve correctly, including the nullable-argument cases, but distinct names make the scope of a destructive operation easier to understand at the call site. Please update the interface, implementation, callers, and tests together.

